PR #4, opened 2017-03-28
Batch copy nodes

robrix:
This PR replaces the directly-bridged API with a batch-copying API. This is much faster, especially for large sources, and generally more convenient to work with as well.


src/bridge.c
```diff
@@ -10,104 +22,48 @@ void ts_document_log_to_stderr(TSDocument *document) {
...
+
+void ts_node_copy_child_nodes(const TSDocument *document, const TSNode *parentNode, Node *outChildNodes, size_t count) {
+  assert(document != NULL);
+  assert(parentNode != NULL);
+  assert(outChildNodes != NULL);
+  assert(count >= 0);
+  uint32_t maxCount = ts_node_child_count(*parentNode);
+  uint32_t max = maxCount <= count ? maxCount : count;
+  for (uint32_t i = 0; i < max; i++) {
+    outChildNodes[i] = ts_node_elaborate(document, ts_node_child(*parentNode, i));
```
robrix: I’m a tiny bit bummed that we can’t `memcpy`, or even better, return a pointer to an already-populated buffer with all of the information in `Node` present. Even with all of the optimizations presently in place, simply copying node buffers using `ts_node_copy_named_child_nodes` (below) takes ~4.6% of our runtime; by contrast, `peek` on `Node` takes ~0.8%.

I’d like to consider alternatives for a future PR, however, as this is already a substantial improvement over our current behaviour.

PR #6, opened 2017-04-04
Unsafe FFI calls

robrix:
`unsafe` means that we know that these will never call back into Haskell, and thus the runtime can skip a bunch of checks that it would otherwise need to use.

This is saving us about half a second in the case I’m testing.


joshvera:
👍

maxbrunsfeld:
💥

PR #7, opened 2017-04-11
Templated production types

robrix:
This PR:

- [x] Exports the language/symbol API.
- [x] Defines a `TemplateHaskell` mechanism for producing a datatype with a constructor for each production in the grammar.
- [x] Defines a mapping from strings onto Haskell constructor names.
- [x] Depends on https://github.com/tree-sitter/tree-sitter/pull/73


robrix:
Ready for review, modulo the dependency on tree-sitter/tree-sitter#73

PR #8, opened 2017-04-12
Allow null document pointers

robrix:
This is intended to ease some experimentation where we don’t need the production name, just the symbol (which is available as of #7).

- [x] Depends on #7.


robrix:
Ready for review.

robrix:
@joshvera: Only when we pass it `nullPtr` (which is now legal).

PR #9, opened 2017-04-13
Classify symbols

robrix:
This PR defines a `Symbol` typeclass, enabling consumers of symbol datatypes to query whether a given symbol is anonymous &c.

- [x] Depends on #8.


robrix:
Ready for review.

PR #10, opened 2017-05-17
Derive Bounded & Ix instances for grammar datatypes.

robrix:
A `Bounded` instance allows us to bounds-check uses of `toEnum` to e.g. deal with `ERROR` productions. An `Ix` instance allows us to perform that bounds-check using the `inRange` method.


robrix:
Ready for review.

robrix:
Ready for review.

PR #15, opened 2017-07-14
Parse errors as maximal symbol

robrix:
This PR places the `ParseError` constructor at the end, and resumes deriving the `Enum` instance, fixing the mismatch between unsigned `TSSymbol` values and the `ParseError` constructor, at the cost of requiring a bounds check on `nodeSymbol` before using `toEnum`.


joshvera:
👍

PR #17, opened 2017-07-23
ghc-8.2.1

robrix:
This PR adds support for ghc-8.2.1 by requiring template-haskell-2.12.0.0 & adding an explicit `DerivClause`.

We should hold off on merging this until we can solve all of our dependencies with that template-haskell version.


.travis.yml
```diff
@@ -1,22 +1,42 @@
...
+- rm -fv cabal.project.local
+- rm -f cabal.project.freeze
+- cabal new-build -w ${HC} --dep -j
 
 script:
-# Build the package, its tests, and its docs and run the tests
-- stack --no-terminal test --haddock --no-haddock-deps
+- rm -rf .ghc.environment.* dist/
+- cabal new-build -w ${HC} ${TEST}
+- cabal new-test -w ${HC} ${TEST}
```
robrix: This all cribs pretty heavily from the `lens` package’s travis config.

src/Text/Parser/TreeSitter/Language.hs
```diff
@@ -38,7 +38,7 @@ mkSymbolDatatype name language = do
 
   Module _ modName <- thisModule
   pure
-    [ DataD [] name [] Nothing (flip NormalC [] . mkName <$> namedSymbols) [ ConT ''Show, ConT ''Enum, ConT ''Eq, ConT ''Ord, ConT ''Bounded, ConT ''Ix ]
+    [ DataD [] name [] Nothing (flip NormalC [] . mkName <$> namedSymbols) [ DerivClause Nothing [ ConT ''Show, ConT ''Enum, ConT ''Eq, ConT ''Ord, ConT ''Bounded, ConT ''Ix ] ]
```
robrix: This is needed because ghc 8.2 adds support for deriving _strategies_; e.g. some instances can be derived by `DeriveAnyClass`, while others might be derived via `GeneralizedNewtypeDeriving`. This can be expressed syntactically, and thus also needs to be expressible in the quasiquoted representation.

src/Text/Parser/TreeSitter/Node.hs
```diff
@@ -75,13 +75,13 @@ instance Storable TSNode where
...
     (o2, ptr) <- peekAdvance ptr
-    (o3, ptr) <- peekAdvance ptr
+    (o3, _) <- peekAdvance ptr
     return $! TSNode p o1 o2 o3
   poke ptr (TSNode p o1 o2 o3) = do
     ptr <- pokeAdvance (castPtr ptr) p
     ptr <- pokeAdvance ptr o1
     ptr <- pokeAdvance ptr o2
-    ptr <- pokeAdvance ptr o3
+    _ <- pokeAdvance ptr o3
```
robrix: Just fixing some warnings while I’m in the neighbourhood.

PR #18, opened 2017-08-01
Upgrade to the newest tree-sitter API

tclem:


src/Text/Parser/TreeSitter/Language.hs
```diff
@@ -33,7 +34,7 @@ class Symbol s where
 mkSymbolDatatype :: Name -> Ptr Language -> Q [Dec]
 mkSymbolDatatype name language = do
   symbols <- (++ [(Regular, "ParseError")]) <$> runIO (languageSymbols language)
-  let namedSymbols = uncurry symbolToName <$> symbols
+  let namedSymbols = nub (uncurry symbolToName <$> symbols)
```
tclem: This is `O(n^2)`... but total number of productions isn't crazy large - maybe in the hundreds.

tclem:
Going to re-cut this...

PR #20, opened 2017-08-10
Name constructors uniquely

rewinfrey:
With `alias` rules we produce symbols of the same name multiple times. This prevents duplicate symbol names from occurring in the named symbols returned from haskell tree sitter for a given tree-sitter grammar.

Duplicate symbols' names are updated with a `'` to maintain uniqueness.

🎩  @robrix for 🍐  and deep dive.


robrix:
Recusing myself from review since I 🍐’d on it; therefore nominated @joshvera 😁

joshvera:
👍

PR #27, opened 2017-10-05
Rename cabal files

rewinfrey:
I started to see the following error messages:

```
cabal file path haskell-tree-sitter/languages/go/go.cabal does not match the package name it defines.
Please rename the file to: tree-sitter-go.cabal
For more information, see: https://github.com/commercialhaskell/stack/issues/317
```

The issue link provided explains that the name of the package should be the same name as the cabal file. I am surprised to see this pop up only now though.


joshvera:
Yeah I saw this too, I think it’s a cabal 2.0 warning.

PR #33, opened 2017-11-03
Add stack files to prevent haskell-tree-sitter from rebuilding

joshvera:
It seems stack needs these in order to prevent rebuilding when running `stack build` multiple times.

### DIFF
--- a/languages/ruby/stack.yaml
+++ b/languages/ruby/stack.yaml
@@ -1,0 +1,8 @@
+flags: {}
+extra-package-dbs: []
+packages:
+- '.'
+- '../../'
+extra-deps:
+- c-storable-deriving-0.1.3
+resolver: nightly-2017-09-10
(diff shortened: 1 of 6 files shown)


languages/ruby/stack.yaml
```diff
@@ -0,0 +1,8 @@
+flags: {}
+extra-package-dbs: []
+packages:
+- '.'
+- '../../'
+extra-deps:
+- c-storable-deriving-0.1.3
```
robrix: I _think_ we shouldn’t need this extra-dep because we don’t depend on c-storable-deriving in the languages.

joshvera: It looks like we need it since the language packages have `haskell-tree-sitter` as a build target.

joshvera:
> Should we add stack.yaml files for Python, Go, JSON, etc?

👍

PR #36, opened 2017-12-19
Bump tree-sitter-go: Add new named nodes

rewinfrey:
This tree-sitter-go bump makes `argument_list`, `field_declaration_list`, `parameter_list`, and `method_spec_list` visible in the parse AST.


joshvera:
👍

PR #37, opened 2018-01-23
++tree-sitter and tree-sitter ruby

tclem:
Bump to pick up https://github.com/tree-sitter/tree-sitter-ruby/commit/6c23e3c7ecb5620ec878fc21fbac91c29d75e1f2 and latest tree-sitter


robrix:
Subsumed by #38.

PR #38, opened 2018-01-23
Remove c-storable-deriving, bump all the things

### LINKED ISSUE #39: Haskell & C types can get out of sync silently
`TSNode`, `TSPoint`, and `Node` are bridged by copying (using `Storable`), but there’s no verification or validation that they’ve got the same shape so they can silently get out of sync.

robrix:
This PR 🔥s the dependency on `c-storable-deriving`, opting to implement `poke` manually instead for the various node types.

Fixes #39.

### DIFF
--- a/haskell-tree-sitter.cabal
+++ b/haskell-tree-sitter.cabal
@@ -37,6 +37,7 @@ library
                      , vendor/tree-sitter/src/runtime/utf16.c
                      , vendor/tree-sitter/externals/utf8proc/utf8proc.c
   cc-options:          -std=c99 -Os
+  ghc-options:         -Wall -fno-warn-name-shadowing -O -j
 
 test-suite haskell-tree-sitter-test
   type:                exitcode-stdio-1.0
--- a/haskell-tree-sitter.cabal
+++ b/haskell-tree-sitter.cabal
@@ -45,6 +45,7 @@ test-suite haskell-tree-sitter-test
   main-is:             Spec.hs
   build-depends:       base
                      , haskell-tree-sitter
+                     , hspec
   ghc-options:         -threaded -rtsopts -with-rtsopts=-N
   default-language:    Haskell2010
 
--- a/haskell-tree-sitter.cabal
+++ b/haskell-tree-sitter.cabal
@@ -39,7 +39,7 @@ library
   cc-options:          -std=c99 -Os
   ghc-options:         -Wall -fno-warn-name-shadowing -O -j
 
-test-suite haskell-tree-sitter-test
+test-suite test
   type:                exitcode-stdio-1.0
   hs-source-dirs:      test
   main-is:             Spec.hs
--- a/haskell-tree-sitter.cabal
+++ b/haskell-tree-sitter.cabal
@@ -16,7 +16,6 @@ cabal-version:       >=1.10
 library
   hs-source-dirs:      src
   build-depends:       base >= 4.7 && < 5
-                     , c-storable-deriving
                      , directory >= 1.3.0.2 && < 1.4
                      , filepath >= 1.4.1.2 && < 1.5
                      , split >= 0.2.3.2 && < 0.3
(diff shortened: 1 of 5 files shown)


src/TreeSitter/Node.hs
```diff
@@ -20,70 +26,107 @@ data Node = Node
...
-
+  deriving (Show, Eq, Generic)
 
 data TSNode = TSNode !(Ptr ()) !Int32 !Int32 !Int32
-  deriving (Show, Eq, Generic, CStorable)
+  deriving (Show, Eq, Generic)
+
+
+-- | 'Struct' is a strict 'Monad' with automatic alignment & advancing, & inferred type.
+newtype Struct a = Struct { runStruct :: forall b . Ptr b -> IO (a, Ptr a) }
```
robrix: I got sick of managing pointers explicitly in the `Storable` instances, so I made a DSL to do it for me. Should not impact performance as the combinators evaluate strictly and are all inlined.

robrix:
This PR has subsumed #37.

src/TreeSitter/Node.hs
```diff
@@ -20,70 +26,107 @@ data Node = Node
...
   , nodeChildCount :: !Int32
   }
-  deriving (Show, Eq, Generic, CStorable)
+  deriving (Show, Eq, Generic)
 
 data TSPoint = TSPoint { pointRow :: !Int32, pointColumn :: !Int32 }
-  deriving (Show, Eq, Generic, CStorable)
+  deriving (Show, Eq, Generic)
 
+data TSNode = TSNode !(Ptr ()) !Word32 !Word32
```
robrix: 1. `TSNode` dropped one element from its array field.
2. The array fields are unsigned, so `Word32` is more appropriate.

src/TreeSitter/Node.hs
```diff
@@ -20,70 +26,107 @@ data Node = Node
...
-pokeAdvance ptr a = do
-  let aligned = alignPtr ptr (alignment (undefined :: a))
+pokeStruct :: Storable a => a -> Struct ()
+pokeStruct a = Struct (\ p -> do
+  let aligned = alignPtr (castPtr p) (alignment a)
   poke aligned a
-  return (castPtr aligned `plusPtr` sizeOf a)
-{-# INLINE pokeAdvance #-}
+  pure ((), castPtr aligned `plusPtr` sizeOf a))
+{-# INLINE pokeStruct #-}
```
robrix: Inlining these means that `Struct` may be slightly faster than the `peekAdvance`/`pokeAdvance` functions which predated them.

haskell-tree-sitter.cabal
```diff
@@ -38,6 +37,7 @@ library
                      , vendor/tree-sitter/src/runtime/utf16.c
                      , vendor/tree-sitter/externals/utf8proc/utf8proc.c
   cc-options:          -std=c99 -Os
+  ghc-options:         -Wall -fno-warn-name-shadowing -O -j
```
tclem: Thank you - this was so noisy.

src/TreeSitter/Node.hs
```diff
@@ -15,78 +20,111 @@ data Node = Node
...
+  deriving (Show, Eq, Generic)
 
-data TSNode = TSNode !(Ptr ()) !Int32 !Int32 !Int32
-  deriving (Show, Eq, Generic, CStorable)
 
-peekAdvance :: forall a b. Storable a => Ptr a -> IO (a, Ptr b)
-peekAdvance ptr = do
-  let aligned = alignPtr ptr (alignment (undefined :: a))
+-- | 'Struct' is a strict 'Monad' with automatic alignment & advancing, & inferred type.
+newtype Struct a = Struct { runStruct :: forall b . Ptr b -> IO (a, Ptr a) }
```
robrix: I briefly experimented with a deep embedding of this DSL with the goal of being able to capture the `alignment` and `sizeOf` operations as well, but it ended up being kind of clumsy to express since peeking had to provide a value before it could evaluate the rest of it. Representing this DSL with an `Applicative` instance but not `Monad` would do nicely, but I was getting off track so I just left it as a shallow embedding.

PR #60, opened 2018-07-23
++tree-sitter and all the languages

tclem:


joshvera:
https://github.com/tree-sitter/haskell-tree-sitter/pull/61 may have subsumed this.

PR #74, opened 2018-11-08
added Demo.hs for existing FFI + DemoPtr.hs for alternate FFI + update to stack lts-12.17

clojj:
…ased FFI

updated to stack lts-12.17

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "C_Cpp.errorSquiggles": "Disabled"
+}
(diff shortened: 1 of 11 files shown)


clojj:
Tree.Ptr is usable, but far from finished... for example, converting an AST to Data.Tree needs to hold on to the TSNode elements somehow (which of course the existing API already provides)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
joshvera: Mind removing this file from the PR?

clojj: shouldn't be in here... sorry, please remove

PR #75, opened 2018-11-15
read the entire tree by using a cursor (Cursor.hs)

clojj:
next TODOs would be
- read the tree into a (newly created) Data.Tree
- add some pretty-printed output (of the Data.Tree)
- add some QuickCheck tests for the tree-walking algorithm

read-tree-sitter-with-cursor

### DIFF
--- a/languages/haskell/examples/DemoReadTreeWithCursor.hs
+++ b/languages/haskell/examples/DemoReadTreeWithCursor.hs
@@ -37,11 +37,4 @@ main = do
 
   withForeignPtr fgnPtr $ \ptr -> do
     ts_ptr_init tree ptr
-    -- when (status /= 0) $ fail "error in init"
-    ts_ptr_goto_first_child ptr
-    ts_ptr_goto_next_sibling ptr
-    ts_ptr_goto_next_sibling ptr
-
-    nodeType <- ts_ptr_current_type ptr
-    theType <- peekCString nodeType
-    print theType
+    readTreeSitter ptr
--- a/src/bridge_ptr.c
+++ b/src/bridge_ptr.c
@@ -17,29 +17,43 @@ void ts_ptr_init(TSTree *tree, tree_sitter_ptr *p)
     TSNode rootNode = ts_tree_root_node(tree);
     assert(rootNode.id != NULL);
     p->rootNode = rootNode;
+    p->cursor = ts_tree_cursor_new(p->rootNode);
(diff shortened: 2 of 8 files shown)


clojj:
Hi,
question regarding future PRs:

as stack does not provide the rosezipper package
http://hackage.haskell.org/package/rosezipper
... I would like to include that package locally.. is that a good idea ?

languages/haskell/examples/DemoReadTreeWithCursor.hs
```diff
@@ -0,0 +1,40 @@
...
+  (str, len) <- newCStringLen source
+
+  tree       <- ts_parser_parse_string parser nullPtr str len
+
+  fgnPtr     <- mallocForeignPtr :: IO (ForeignPtr TreeSitter_Ptr)
+  addForeignPtrFinalizer funptr_ts_ptr_free fgnPtr
+
+  withForeignPtr fgnPtr $ \ptr -> do
+    ts_ptr_init tree ptr
+    readTreeSitter ptr
```
clojj: in the next PR, I will construct a Data.Tree here (instead of just printing the nodes)

src/bridge_ptr.c
```diff
@@ -17,29 +17,43 @@ void ts_ptr_init(TSTree *tree, tree_sitter_ptr *p)
...
 }
 
-void ts_ptr_goto_first_child(tree_sitter_ptr *p)
+bool ts_ptr_goto_first_child(tree_sitter_ptr *p)
 {
-    p->cursor = ts_tree_cursor_new(p->rootNode);
     if (ts_tree_cursor_goto_first_child(&p->cursor))
     {
-        debugPrintCurrentNode(p);
+        // debugPrintCurrentNode(p);
```
clojj: these empty blocks will have node related mappings

clojj:
closing this.. it is a bit of a mess
I will soon bring a new PR that should be much better

joshvera:
> as stack does not provide the rosezipper package
http://hackage.haskell.org/package/rosezipper
... I would like to include that package locally.. is that a good idea ?

@clojj yeah I think we can add it to stack with a git URL. Looks like you’ve made some additions at https://github.com/yav/haskell-zipper.

clojj:
@joshvera 
yes, and I asked the maintainer to publish rosezipper on stackage, we'll see

PR #76, opened 2018-11-16
enable new cursor-level access

clojj:
demo here:
/languages/haskell/examples/DemoReadTreeWithCursor.hs
For now, I crudely copy-pasted some code from Node.hs (Struct etc.), this should be extracted to a new common module, what do you think ?


clojj:
put Struct in separate module

clojj:
I forgot to add the local rosezipper package (haskell-zipper)... it's not yet on stackage

clojj:
tree-sitter-cursor

clojj:
closing this for another PR

PR #77, opened 2018-11-21
cursor API + demo and hspec-tests

clojj:


clojj:
- package rosezipper should be on stackage soon... so the submodule can be removed
- package loc is just a local package for now.. it's not on stackage yet

robrix:
This is very far out of date, and we have `TreeSitter.Cursor` now, so I’m going to close this out for the time being.

PR #84, opened 2019-03-18
Update Tree-sitter, expose new timeout API

maxbrunsfeld:
Now that https://github.com/tree-sitter/tree-sitter/pull/301 is done, we have a simpler way to impose timeouts on parsing.

This PR updates the Tree-sitter submodule and exposes the new timeout functions. I've also removed the `set_enabled` functions, because those underlying Tree-sitter functions were changed in https://github.com/tree-sitter/tree-sitter/pull/306.

### DIFF
--- a/haskell-tree-sitter.cabal
+++ b/haskell-tree-sitter.cabal
@@ -26,21 +26,14 @@ library
                      , TreeSitter.Node
                      , TreeSitter.Tree
                      , TreeSitter.Ptr
-  Include-dirs:        includes                     
-  Includes:            tree_sitter_ptr.h                     
+  Include-dirs:        includes
+                     , vendor/tree-sitter/lib/include
+                     , vendor/tree-sitter/lib/src
+                     , vendor/tree-sitter/lib/utf8proc
+  Includes:            tree_sitter_ptr.h
   c-sources:           src/bridge.c
                      , src/bridge_ptr.c
-                     , vendor/tree-sitter/externals/utf8proc/utf8proc.c
-                     , vendor/tree-sitter/src/runtime/get_changed_ranges.c
-                     , vendor/tree-sitter/src/runtime/language.c
-                     , vendor/tree-sitter/src/runtime/lexer.c
-                     , vendor/tree-sitter/src/runtime/node.c
-                     , vendor/tree-sitter/src/runtime/parser.c
-                     , vendor/tree-sitter/src/runtime/stack.c
-                     , vendor/tree-sitter/src/runtime/subtree.c
-                     , vendor/tree-sitter/src/runtime/tree.c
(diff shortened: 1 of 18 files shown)


haskell-tree-sitter.cabal
```diff
@@ -26,21 +26,14 @@ library
...
-                     , vendor/tree-sitter/src/runtime/language.c
-                     , vendor/tree-sitter/src/runtime/lexer.c
-                     , vendor/tree-sitter/src/runtime/node.c
-                     , vendor/tree-sitter/src/runtime/parser.c
-                     , vendor/tree-sitter/src/runtime/stack.c
-                     , vendor/tree-sitter/src/runtime/subtree.c
-                     , vendor/tree-sitter/src/runtime/tree.c
-                     , vendor/tree-sitter/src/runtime/tree_cursor.c
-                     , vendor/tree-sitter/src/runtime/utf16.c
+                     , vendor/tree-sitter/lib/src/lib.c
```
maxbrunsfeld: There's now a single 'amalgamation build' file, `lib.c`, so we don't have to list out all of the source files here.

haskell-tree-sitter.cabal
```diff
@@ -26,21 +26,14 @@ library
                      , TreeSitter.Node
                      , TreeSitter.Tree
                      , TreeSitter.Ptr
-  Include-dirs:        includes                     
-  Includes:            tree_sitter_ptr.h                     
+  Include-dirs:        includes
+                     , vendor/tree-sitter/lib/include
+                     , vendor/tree-sitter/lib/src
+                     , vendor/tree-sitter/lib/utf8proc
```
maxbrunsfeld: I couldn't tell how these include directories were specified *before*; maybe `cabal` or the C compiler was just somehow inferring them?

test/Spec.hs
```diff
@@ -27,6 +28,16 @@ main = hspec $ do
...
 
+  describe "Parser" $ do
+    it "stores a timeout value" $ do
+      parser <- ts_parser_new
+      timeout <- ts_parser_timeout_micros parser
+      timeout `shouldBe` 0
+      ts_parser_set_timeout_micros parser 1000
+      timeout <- ts_parser_timeout_micros parser
+      timeout `shouldBe` 1000
+      ts_parser_delete parser
```
maxbrunsfeld: This test doesn't do anything interesting *with* the timeout (like parsing and letting it expire) but it seemed worth checking that the numeric argument gets wired through correctly.

haskell-tree-sitter.cabal
maxbrunsfeld: Oh, nevermind. We were specifying the full path to the header file in each `foreign import` declaration.

PR #86, opened 2019-05-06
Move th-syntax contents

aymannadeem:
Moving over the Template Haskell bits prototyped in `th-syntax` that produce datatypes, and starting to work on producing values of those datatypes using the FFI as per [the plan outlined here](https://github.com/github/th-syntax/issues/32).


aymannadeem:
I pulled master and some submoduled weirdness crept in. Will fix, ignore for now.

aymannadeem:
Made some good progress today defining instances for basic scenarios (ex., products, sums, leaves) and pairing with @robrix on [importing data types via the Cursor API](https://github.com/tree-sitter/haskell-tree-sitter/pull/88). 

next steps:
- [ ] `Alternative` instance for `MaybeC`
- [ ] `Monad`/`MonadIO` instances for `MaybeC`
- [ ] Continue making basic instances (ex., lists)

aymannadeem:
Addressed #93 with https://github.com/tree-sitter/haskell-tree-sitter/pull/86/commits/f3b56c51a7ebbf5c8a7cf78c9d81279226bf1f94.

src/TreeSitter/Ptr.hs
```diff
@@ -1,26 +1,19 @@
...
 
 import TreeSitter.Tree
 
+foreign import ccall unsafe "src/bridge.c sizeof_tsnode" sizeof_tsnode :: CSize
+foreign import ccall unsafe "src/bridge.c sizeof_tstreecursor" sizeof_tstreecursor :: CSize
 
-data TreeSitter_Ptr
+newtype TreeSitter_Ptr = TreeSitter_Ptr ()
 
 instance Storable TreeSitter_Ptr where
```
aymannadeem: In order to work and successfully load the repl, I had to comment out `TreeSitter.Ptr` in the cabal file.

@patrickt pointed out that the reason that was failing is because that file used a custom C preprocessor c2hs, and he issued this fix.

PR #90, opened 2019-05-15
Generic building

robrix:
This PR defines a generic mechanism for building AST data from maps of field names to `TSNode`s; combined with `Importing` (which we should maybe rename?) this will make for a nice way to use the cursor API to copy named fields over the FFI ✨

**To-do:**

- [ ] Use the symbol to select the right type to decode with. 🎩 @maxbrunsfeld for the excellent suggestion.


aymannadeem:
Yes to the rename. 😬

src/TreeSitter/Importing.hs
```diff
@@ -1,116 +1,146 @@
...
-  import' = do
-    node <- peekNode
-    bytestring <- ask
-    let start = fromIntegral (nodeStartByte node)
-        end = fromIntegral (nodeEndByte node)
-    pure (decodeUtf8 (slice start end bytestring))
+instance (Building a, Building b) => Building (Either a b) where
+  buildNode = Left <$> buildNode <|> Right <$> buildNode
 
+instance Building a => Building [a] where
```
aymannadeem: #### Why we need both `Building` and `GBuilding`

We have two classes: `Building` and `GBuilding`. This is because generic behaviors are different than what we get non-generically, and in the case of `Text`, `Maybe`, `[]`, `Either`—we actually preference doing things non-generically. Why? Well, since `[]` is a sum, the generic behaviour for `:+:` would be invoked and expect that we’d have repetitions represented in the parse tree as right-nested singly-linked lists (ex., `(a (b (c (d…))))`) rather than as just consecutive sibling nodes (ex., `(a b c ...d)`, which is what our trees have). We want to match the latter.

src/TreeSitter/Importing.hs
```diff
@@ -1,116 +1,146 @@
...
-    b <- import' @b
-    pure (a, b)
+import           GHC.Generics
+import           TreeSitter.Cursor as TS
+import           TreeSitter.Language as TS
+import           TreeSitter.Node as TS
+import           TreeSitter.Parser as TS
+import           TreeSitter.Tree as TS
+
+parseByteString :: Building t => Ptr TS.Language -> ByteString -> IO (Maybe t)
```
aymannadeem: We take source code in the form of a bytestring and produce AST values represented by `IO Maybe`

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +158,56 @@ instance Monad m => Monad (MaybeC m) where
...
     case a' of
       Nothing -> pure Nothing
-      Just a -> runMaybeC $ f a
+      Just a  -> runMaybeC $ f a
 
 instance MonadIO m => MonadIO (MaybeC m) where
   liftIO = MaybeC . fmap Just . liftIO
+
+
+newtype FieldName = FieldName { getFieldName :: String }
```
aymannadeem: Examples of field names are left, right, operator, or for an if statement, something like `condition`, `consequence`, `alternative`. In the `CodeGen.Deserialize` code, field name is used to name the record fields of product data types. 

The generic building code uses the field name for the current node to select which record field should build with the sub-tree.

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +158,56 @@ instance Monad m => Monad (MaybeC m) where
...
+instance (GBuilding f, Selector c) => GBuilding (M1 S c f) where
+  gbuildNode fields =
+    case Map.lookup (FieldName (selName @c undefined)) fields of
+      Just node -> do
+        goto (nodeTSNode node)
+        M1 <$> gbuildNode fields
+      Nothing -> M1 <$> gbuildEmpty
+  gbuildEmpty = M1 <$> gbuildEmpty
+
+instance Building c => GBuilding (K1 i c) where
```
aymannadeem: This GBuilding instance for constant `K1` calls into the `Building` instance for the field type, which is what lets us treat our special cases above differently (`Maybe`, `[]`, etc.) wherever they occur in the tree.

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +158,56 @@ instance Monad m => Monad (MaybeC m) where
...
+
+-- Build an empty node of given type
+  gbuildEmpty :: Alternative m => m (f a)
+
+instance (GBuilding f, GBuilding g) => GBuilding (f :*: g) where
+  gbuildNode fields = (:*:) <$> gbuildNode @f fields <*> gbuildNode @g fields
+  gbuildEmpty = (:*:) <$> gbuildEmpty @f <*> gbuildEmpty @g
+
+instance (GBuilding f, GBuilding g) => GBuilding (f :+: g) where
+  gbuildNode fields = L1 <$> gbuildNode @f fields <|> R1 <$> gbuildNode @g fields
```
aymannadeem: We will eventually wanna lose the `Alternation` here in favor of committed choice.

src/TreeSitter/Importing.hs
```diff
@@ -1,116 +1,146 @@
...
+import           TreeSitter.Tree as TS
+
+parseByteString :: Building t => Ptr TS.Language -> ByteString -> IO (Maybe t)
+parseByteString language bytestring = withParser language $ \ parser -> withParseTree parser bytestring $ \ treePtr ->
+  if treePtr == nullPtr then
+    pure Nothing
+  else
+    withRootNode treePtr $ \ rootPtr ->
+      withCursor (castPtr rootPtr) $ \ cursor ->
+        runMaybeC (runM (runReader cursor (runReader bytestring buildNode)))
```
aymannadeem: We'll probably replace `runMaybeC` with error

src/TreeSitter/Importing.hs
```diff
@@ -1,116 +1,147 @@
...
+
+-- | Building is the process of iterating over tree-sitter’s parse trees using its tree cursor API and producing Haskell ASTs for the relevant nodes.
+--
+--   Datatypes which can be constructed from tree-sitter parse trees may use the default definition of 'buildNode' providing that they have a suitable 'Generic' instance.
+class Building a where
+  buildNode :: (Alternative m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m a
+  default buildNode :: (Alternative m, Carrier sig m, GBuilding (Rep a), Generic a, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m a
+  buildNode = to <$> gbuildNode
```
aymannadeem: `gbuildNode` is what we use to match the node, and then fmapping `to :: Rep a x -> a` over it just allows us to wrap that representation into the `a` we want.

src/TreeSitter/Importing.hs
```diff
@@ -1,116 +1,147 @@
...
+  else
+    withRootNode treePtr $ \ rootPtr ->
+      withCursor (castPtr rootPtr) $ \ cursor ->
+        runMaybeC (runM (runReader cursor (runReader bytestring buildNode)))
+
+-- | Building is the process of iterating over tree-sitter’s parse trees using its tree cursor API and producing Haskell ASTs for the relevant nodes.
+--
+--   Datatypes which can be constructed from tree-sitter parse trees may use the default definition of 'buildNode' providing that they have a suitable 'Generic' instance.
+class Building a where
+  buildNode :: (Alternative m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m a
```
aymannadeem: Probably worth mentioning why we use fused effects. `Building`, formerly known as `Importing`, used the original method definition `import' :: Node -> type'`. It would ingest a `Node` and use that information to spit out types, for the purpose of creating a statically typed tree.

However, we need fused-effects to pass the `bytestring` around rather than doing so manually. That would require that `ByteString` would be a parameter to every function, and if you need to make modifications to that `ByteString` that other states need to see, returning it as a value as well when necessary. This can be error-prone and tedious, so we substitute that entire approach with the use of a `Reader` monad. A `Reader` monad is semantically equivalent to passing around a value as an argument. 

`Node` alone doesn't (and shouldn't) know about `ByteString`. And we need both in order to convert something (like a leaf node) into something else (like `Text`). `ByteString` tells us which portion of the `Node` to cut down and convert.

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +159,71 @@ instance Monad m => Monad (MaybeC m) where
...
+  gbuildSumNode = L1 <$> gbuildSumNode @f <|> R1 <$> gbuildSumNode @g
+
+class GBuildingProduct f where
+  gbuildProductNode :: (Alternative m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => Map.Map FieldName Node -> m (f a)
+-- Product structure
+instance (GBuildingProduct f, GBuildingProduct g) => GBuildingProduct (f :*: g) where
+  gbuildProductNode fields = (:*:) <$> gbuildProductNode @f fields <*> gbuildProductNode @g fields
+
+-- Contents of product types (ie., the leaves of the product tree)
+instance (Building k, Selector c) => GBuildingProduct (M1 S c (K1 i k)) where
```
aymannadeem: By specializing this to use `K1`, we could eliminate the separate instance for `K1` that we had originally:

```
instance Building c => GBuildingProduct (K1 i c) where
  gbuildProductNode _ = K1 <$> buildNode
  gbuildProductEmpty = K1 <$> buildEmpty
```

Additionally, we can axe `U1` since `U1` is only ever used as the contents of constructor meta-data type, which only happens when you have something like `data Foo = Foo`

```
instance GBuilding U1 where
  gbuildNode _ = pure U1
  gbuildEmpty = empty
```

src/TreeSitter/Importing.hs
```diff
@@ -203,17 +203,11 @@ class GBuildingProduct f where
...
-
-instance Building c => GBuilding (K1 i c) where
-  gbuildNode _ = K1 <$> buildNode
-  gbuildEmpty = K1 <$> buildEmpty
-
-instance GBuilding U1 where
-  gbuildNode _ = pure U1
-  gbuildEmpty = empty
+        M1 . K1 <$> buildNode
+      Nothing -> M1 . K1 <$> buildEmpty
```
aymannadeem: This will only happen for `K1` so we don't need to abstract beyond that.

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +159,71 @@ instance Monad m => Monad (MaybeC m) where
...
+  gbuildSumNode :: (Alternative m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m (f a)
+-- we'd only build the map when we know we're looking at a product
+
+instance Building k => GBuildingSum (M1 C c (M1 S s (K1 i k))) where
+  gbuildSumNode = M1 . M1 . K1 <$> buildNode
+
+instance (GBuildingSum f, GBuildingSum g) => GBuildingSum (f :+: g) where
+  gbuildSumNode = L1 <$> gbuildSumNode @f <|> R1 <$> gbuildSumNode @g
+
+class GBuildingProduct f where
```
aymannadeem: By separating this out from the original `GBuilding` class, we'd only build the map when we know we're looking at a product.

src/TreeSitter/Importing.hs
```diff
@@ -128,7 +159,71 @@ instance Monad m => Monad (MaybeC m) where
...
+
+class GBuildingSum f where
+  gbuildSumNode :: (Alternative m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m (f a)
+-- we'd only build the map when we know we're looking at a product
+
+instance Building k => GBuildingSum (M1 C c (M1 S s (K1 i k))) where
+  gbuildSumNode = M1 . M1 . K1 <$> buildNode
+
+instance (GBuildingSum f, GBuildingSum g) => GBuildingSum (f :+: g) where
+  gbuildSumNode = L1 <$> gbuildSumNode @f <|> R1 <$> gbuildSumNode @g
```
aymannadeem: We need to represent this with a committed choice between branches instead of alternation/backtracking.

src/TreeSitter/Parser.hs
```diff
@@ -28,7 +28,7 @@ withParser language action = Exc.bracket
   ts_parser_new
   ts_parser_delete
   $ \ parser -> do
-    ts_parser_set_language parser language
+    _ <- ts_parser_set_language parser language
```
robrix: We should probably actually change the signature of this function to return in `Maybe` based on whether we get `True` or `False` back from this function, but that doesn’t necessarily happen in this PR; feel free to file an issue and punt 👍

src/TreeSitter/Importing.hs
```diff
@@ -35,54 +36,57 @@ import           TreeSitter.Node as TS
...
-        bytestring <- ask
-        let start = fromIntegral (nodeStartByte node')
-            end = fromIntegral (nodeEndByte node')
-        pure (decodeUtf8 (slice start end bytestring))
-      _ -> empty
+  buildEmpty :: MonadFail m => m a
+  buildEmpty = fail "expected a node" -- TODO: log what type actually expected the node / what `a` is
 
-instance Building a => Building (Maybe a) where
+instance Building symbol a => Building symbol (Maybe a) where
```
robrix: Were we still planning on backing the introduction of the `symbol` parameter out?

aymannadeem: Yes! Just wanted to push these changes to have breadcrumbs for myself and know what we've tried.

PR #116, opened 2019-06-05
Re-/de-bump the submodules

robrix:
We accidentally bumped most of the language submodules in #86 (instead of just Python, which we’d intended). This PR bumps them all to `master`, except TypeScript, which it resets back to pre-#86.

- [x] ~~Depends on #115.~~ Nope.


robrix:
Note that alternatively, we could reset all of the submodules except `tree-sitter-python` to their pre-#86 SHAs.

robrix:
(Let’s hold off on merging til I know this didn’t screw things up worse 😁)

robrix:
Confirmed that this is working against https://github.com/github/semantic/pull/95 ✨

PR #144, opened 2019-06-20
Generic un-marshaling

aymannadeem:
This PR defines a generic un-marshaling mechanism to build AST data, using meta-programming to automate the [Assignment](https://github.com/github/semantic/blob/master/docs/assignment.md) component of our language support system. 

We’ve successfully demonstrated that we can replace a la carte syntaxes with datatypes which are auto-generated at compile-time [via Template Haskell](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/src/CodeGen/GenerateSyntax.hs). While that provides us with the exhaustive set of datatypes for a given language being mapped from the representation of the `tree-sitter` `node-types.json` file, we also need to parse source code into their correct AST representation.

This PR accomplishes that by defining:

1. **`parseByteString`:** parse source code and generate the ASTs.
2. **`Building`:** non-generic class and public interface that allows us to iterate over the tree-sitter’s parse tree using its tree cursor API and producing Haskell ASTs for relevant nodes. 
3. **`GBuilding`:** Generic construction of ASTs.

#### Why do we have Generic and Non-Generic representations?

We have a Building class and several generic classes, and the reason is because generic behaviors are different than what we get non-generically, and in the case of `Text`, `Maybe`, `[]`, `Either`—we actually preference doing things non-generically. Why? Well, since `[]` is a sum, the generic behavior for `:+:` would be invoked and expect that we’d have repetitions represented in the parse tree as right-nested singly-linked lists `(ex., (a (b (c (d…)))))` rather than as just consecutive sibling nodes `(ex., (a b c ...d)`, which is what our trees have). We want to match the latter.


languages/python/TreeSitter/Python/AST.hs
```diff
@@ -1,15 +1,18 @@
-{-# LANGUAGE DeriveAnyClass, DeriveGeneric, DuplicateRecordFields, TemplateHaskell #-}
+{-# OPTIONS_GHC -ddump-splices #-}
```
robrix: 🔥

languages/python/tree-sitter-python.cabal
```diff
@@ -35,6 +35,18 @@ library tree-sitter-python-internal
   c-sources:           vendor/tree-sitter-python/src/parser.c
                      , vendor/tree-sitter-python/src/scanner.cc
   extra-libraries:     stdc++
+  ghc-options:         -Weverything -Wno-missing-local-signatures -Wno-missing-import-lists -Wno-implicit-prelude -Wno-safe -Wno-unsafe -Wno-name-shadowing -Wno-monomorphism-restriction -Wno-missed-specialisations -Wno-all-missed-specialisations -Wno-missing-export-lists -ddump-splices
+  if (impl(ghc >= 8.6))
+    ghc-options:       -Wno-star-is-type
+
+executable python
```
robrix: We can 🔥 this `executable` stanza.

languages/python/app/Main.hs
```diff
@@ -0,0 +1,13 @@
...
+import qualified Data.ByteString as B
+import qualified TreeSitter.Importing as TS
+import TreeSitter.Python
+
+parseByteString :: B.ByteString -> IO (Maybe Module)
+parseByteString = TS.parseByteString tree_sitter_python
+
+main :: IO ()
+main = do
+  parseByteString "from __future__ import absolute_import" >>= print
```
robrix: We can 🔥 this file along with the `python` executable in the tree-sitter-python.cabal file.

languages/python/TreeSitter/Python/AST.hs
```diff
@@ -20,4 +23,4 @@ $(do
   pwd             <- runIO getCurrentDirectory
   let invocationRelativePath = takeDirectory (pwd </> currentFilename) </> "../../vendor/tree-sitter-python/src/node-types.json"
   input <- runIO (eitherDecodeFileStrict' invocationRelativePath)
-  either fail (traverse datatypeForConstructors) input)
+  either fail (fmap (concat @[]) . traverse (datatypeForConstructors Grammar.tree_sitter_python)) input)
```
robrix: We’re going to eventually have AST modules for each supported language, each of which need to do this whole flow. I think we should therefore encapsulate the whole thing into something like:

```haskell
astDeclarationsForLanguage :: Ptr Language -> FilePath -> Q [Dec]
```

which we’d call in this module (and other AST modules).

src/CodeGen/GenerateSyntax.hs
```diff
@@ -13,32 +13,62 @@ module CodeGen.GenerateSyntax
...
+import GHC.Generics hiding (Constructor, Datatype)
+import Foreign.Ptr
+import qualified TreeSitter.Language as TS
+import Foreign.C.String
+import Data.Proxy
 
 -- Auto-generate Haskell datatypes for sums, products and leaf types
-datatypeForConstructors :: MkDatatype -> Q Dec
-datatypeForConstructors (SumType (DatatypeName datatypeName) named subtypes) = do
+datatypeForConstructors :: Ptr TS.Language -> MkDatatype -> Q [Dec]
```
robrix: 1. I think we can probably rename this. `datatypeForConstructors` doesn’t really reflect either what it takes or what it does.

2. This would probably be the place to define an `astDeclarationsForLanguage` or similar function (see above).

src/CodeGen/GenerateSyntax.hs
```diff
@@ -108,7 +138,7 @@ addTickIfNecessary s
 
 -- | Convert snake_case string to CamelCase Name
 toName :: String -> Name
-toName = mkName . toCamelCase
+toName = mkName . addTickIfNecessary . toCamelCase
```
robrix: I don’t think we need to do `addTickIfNecessary` here, since `toCamelCase` does `initUpper` and thus the name won’t ever be one of the keywords (`type`, `module`, `data`) that `addTickIfNecessary` was intended to deal with.

By contrast, we use `addTickIfNecessary` in `toVarBangType`, and that seems correct to me.

src/CodeGen/GenerateSyntax.hs
```diff
@@ -108,7 +138,7 @@ addTickIfNecessary s
...
 
 -- | Prepend "Anonymous" to named node when false, otherwise use regular toName
 toName' :: MkNamed -> String -> Name
```
robrix: 💡 We could maybe consider refactoring this into something which operates on a `MkType`, and then introduce a function of type `MkDatatype -> MkType` to deal with the uses in `datatypeForConstructors`, above. Not a blocker for this PR, just a thought.

src/CodeGen/GenerateSyntax.hs
```diff
@@ -13,32 +13,62 @@ module CodeGen.GenerateSyntax
...
-  pure $ NewtypeD [] name [] Nothing con [ DerivClause Nothing [ ConT ''Eq, ConT ''Ord, ConT ''Show ] ]
+  result <- symbolMatchingInstance language name datatypeName
+  pure $ NewtypeD [] name [] Nothing con [ DerivClause Nothing [ ConT ''TS.Building, ConT ''Eq, ConT ''Ord, ConT ''Show, ConT ''Generic ] ]:result
+
+-- | Create TH-generated SymbolMatching instances for sums, products, leaves
+symbolMatchingInstance :: Ptr TS.Language -> Name -> String -> Q [Dec]
+symbolMatchingInstance language name str = do
+  tsSymbol <- runIO $ withCString str (pure . TS.ts_language_symbol_for_name language)
+  tsSymbolType <- pure . toEnum $ TS.ts_language_symbol_type language tsSymbol
+  [d|instance TS.SymbolMatching $(conT name) where
```
robrix: I was really glad to learn that you can indeed do this sort of thing!

src/CodeGen/GenerateSyntax.hs
```diff
@@ -13,32 +13,62 @@ module CodeGen.GenerateSyntax
...
   con <- toConLeaf named (DatatypeName datatypeName)
-  pure $ NewtypeD [] name [] Nothing con [ DerivClause Nothing [ ConT ''Eq, ConT ''Ord, ConT ''Show ] ]
+  result <- symbolMatchingInstance language name datatypeName
+  pure $ NewtypeD [] name [] Nothing con [ DerivClause Nothing [ ConT ''TS.Building, ConT ''Eq, ConT ''Ord, ConT ''Show, ConT ''Generic ] ]:result
+
+-- | Create TH-generated SymbolMatching instances for sums, products, leaves
+symbolMatchingInstance :: Ptr TS.Language -> Name -> String -> Q [Dec]
+symbolMatchingInstance language name str = do
+  tsSymbol <- runIO $ withCString str (pure . TS.ts_language_symbol_for_name language)
+  tsSymbolType <- pure . toEnum $ TS.ts_language_symbol_type language tsSymbol
```
robrix:
```suggestion
  let tsSymbolType = toEnum $ TS.ts_language_symbol_type language tsSymbol
```

src/CodeGen/GenerateSyntax.hs
```diff
@@ -13,32 +13,62 @@ module CodeGen.GenerateSyntax
...
+
+symbolMatchingInstanceForSums ::  Ptr TS.Language -> Name -> [MkType] -> Q [Dec]
+symbolMatchingInstanceForSums _ name subtypes =
+  [d|instance TS.SymbolMatching $(conT name) where
+      showFailure _ node = "Expected " <> $(litE (stringL (show (map extractn subtypes)))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
+      symbolMatch _ = $(foldr1 mkOr (perMkType `map` subtypes)) |]
+  where perMkType (MkType (DatatypeName n) named) = [e|TS.symbolMatch (Proxy :: Proxy $(conT (toName' named n))) |]
+        mkOr lhs rhs = [e| (||) <$> $(lhs) <*> $(rhs) |]
+        extractn (MkType (DatatypeName n) Named) = toCamelCase n
+        extractn (MkType (DatatypeName n) Anonymous) = "Anonymous" <> toCamelCase n
```
robrix: 💡 It might be worth factoring `extractn` out as e.g. `typeName :: MkType -> String`; then `toName'` could actually be replaced with `mkName . typeName` or something like that.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+    withRootNode treePtr $ \ rootPtr ->
+      withCursor (castPtr rootPtr) $ \ cursor ->
+        runM (runFail (runReader cursor (runReader bytestring buildNode)))
 
-data Number = Number
-      deriving (Eq, Ord, Show)
+-- | Building is the process of iterating over tree-sitter’s parse trees using its tree cursor API and producing Haskell ASTs for the relevant nodes.
+--
+--   Datatypes which can be constructed from tree-sitter parse trees may use the default definition of 'buildNode' providing that they have a suitable 'Generic' instance.
+class Building a where
```
robrix: If you’re settled on “unmarshalling” then let‘s rename all the “build” things to “unmarshal” things; then we can close #113.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
--- WIP: Do not depend on this yet!
-{-# LANGUAGE DeriveFunctor, FlexibleContexts, GeneralizedNewtypeDeriving, ScopedTypeVariables, TypeApplications #-}
-module TreeSitter.Importing where
+{-# LANGUAGE DefaultSignatures, DeriveFunctor, FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving,
+             ScopedTypeVariables, TypeApplications, TypeOperators, PolyKinds #-}
```
robrix: Could you please alphabetize this?

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+instance Building a => Building (Maybe a) where
+  buildNode = Just <$> buildNode
+  buildEmpty = pure Nothing
 
-      let release t
-            | t == nullPtr = pure ()
-            | otherwise = ts_tree_delete t
+instance (Building a, Building b, SymbolMatching a, SymbolMatching b) => Building (Either a b) where
+  buildNode = do
+      currentNode <- peekNode
```
robrix: Looks like we’ve got an extra level of indentation going on in this `do` block.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+      (lhsSymbolMatch, rhsSymbolMatch, currentNode) <- case currentNode of
+        Just node -> pure (symbolMatch (Proxy @a) node, symbolMatch (Proxy @b) node, node)
+        Nothing -> fail "expected a node; didn't get one"
```
robrix: We could simplify this logic (and the similar logic over `:+:`s in the generic counterpart) by changing `showFailure` to take `Maybe Node` instead of `Node`. Then we could do `fail $ showFailure (Proxy @(Either a b)) currentNode` on this line, and on the line above, we wouldn’t need to return `node` as part of the triple since the later calls to `showFailure` could use the original binding of `currentNode :: Maybe Node` from the first line of the `do`-block.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+      if lhsSymbolMatch -- FIXME: report error
+        then Left <$> buildNode @a
+        else if rhsSymbolMatch
+          then Right <$> buildNode @b
+          else fail $ showFailure (Proxy @a) currentNode `sep` showFailure (Proxy @b) currentNode -- TODO: do the toEnum nodeSymbol stuff for the current node to show the symbol name
```
robrix: 1. This `TODO` is obsolete.

2. We could simplify this by using the `SymbolMatching` instance for `Either` instead.

```suggestion
          else fail $ showFailure (Proxy @(Either a b)) currentNode
```

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
-  (ts_tree_cursor_delete cursor)
-  (action cursor)
+class SymbolMatching a where
+  symbolMatch :: Proxy a -> Node -> Bool
 
-instance Importing a => Importing [a] where
-  import' = push $ do
-    a <- import' @a
-    pure [a]
+  -- | Return the String describing the type received
```
robrix: Is there a clearer way to describe the purpose and intent of this method?

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+
+
+-- | Generic construction of ASTs from a 'Map.Map' of named fields.
+--
+--   Product types (specifically, record types) are constructed by looking up the node for each corresponding field name in the map, moving the cursor to it, and then invoking 'buildNode' to construct the value for that field. Leaf types are constructed as a special case of product types.
+--
+--   Sum types are constructed by attempting to build each constructor nondeterministically. This should instead use the current node’s symbol to select the corresponding constructor deterministically.
+class GBuilding f where
+  gbuildNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m (f a)
+-- we'd only build the map when we know we're looking at a product
```
robrix:
```suggestion
```

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+
+
+instance GBuilding f => GBuilding (M1 D c f) where
+  gbuildNode = M1 <$> gbuildNode -- current node, not first child like above in original Building definition
```
robrix:
```suggestion
  gbuildNode = M1 <$> gbuildNode
```

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+  -- gSymbolMatch _ = gSymbolMatch (Proxy @f)
```
robrix:
```suggestion
```

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+
+instance GBuilding f => GBuilding (M1 C c f) where
+  gbuildNode = M1 <$> gbuildNode
+
+-- Possibly for anonymous leaf nodes:
```
robrix:
```suggestion
-- For anonymous leaf nodes:
```

src/TreeSitter/Language.hs
```diff
@@ -23,6 +23,7 @@ newtype Language = Language ()
 foreign import ccall unsafe "ts_language_symbol_count" ts_language_symbol_count :: Ptr Language -> Word32
 foreign import ccall unsafe "ts_language_symbol_name" ts_language_symbol_name :: Ptr Language -> TSSymbol -> CString
 foreign import ccall unsafe "ts_language_symbol_type" ts_language_symbol_type :: Ptr Language -> TSSymbol -> Int
+foreign import ccall unsafe "ts_language_symbol_for_name" ts_language_symbol_for_name :: Ptr Language -> CString -> TSSymbol
```
robrix: We should probably double-check with @maxbrunsfeld whether these are actually referentially transparent or not.

languages/python/tree-sitter-python.cabal
```diff
@@ -35,6 +35,18 @@ library tree-sitter-python-internal
   c-sources:           vendor/tree-sitter-python/src/parser.c
                      , vendor/tree-sitter-python/src/scanner.cc
   extra-libraries:     stdc++
+  ghc-options:         -Weverything -Wno-missing-local-signatures -Wno-missing-import-lists -Wno-implicit-prelude -Wno-safe -Wno-unsafe -Wno-name-shadowing -Wno-monomorphism-restriction -Wno-missed-specialisations -Wno-all-missed-specialisations -Wno-missing-export-lists -ddump-splices
```
patrickt: Kill the `-ddump-splices` on this line here, too.

src/CodeGen/GenerateSyntax.hs
```diff
@@ -13,32 +13,62 @@ module CodeGen.GenerateSyntax
...
-datatypeForConstructors :: MkDatatype -> Q Dec
-datatypeForConstructors (SumType (DatatypeName datatypeName) named subtypes) = do
+datatypeForConstructors :: Ptr TS.Language -> MkDatatype -> Q [Dec]
+datatypeForConstructors language (SumType (DatatypeName datatypeName) named subtypes) = do
   let name = toName' named datatypeName
   cons <- traverse (toSumCon datatypeName) subtypes
-  pure $ DataD [] name [] Nothing cons [ DerivClause Nothing [ ConT ''Eq, ConT ''Ord, ConT ''Show ] ]
-datatypeForConstructors (ProductType (DatatypeName datatypeName) named fields) = do
+  result <- symbolMatchingInstanceForSums language name subtypes
+  pure $ DataD [] name [] Nothing cons [ DerivClause Nothing [ ConT ''TS.Building, ConT ''Eq, ConT ''Generic, ConT ''Ord, ConT ''Show ] ]:result
```
patrickt: Can we reduce the duplication in these stanzas? I would probably start with pulling out the `[ConT ''TS.Building, etc.]` into

```
standardASTDerivations :: DerivClause
standardASTDerivations = DerivClause Nothing [ ConT ''TS.Building, ConT ''Eq, ConT ''Generic, ConT ''Ord, ConT ''Show ] ]
```

But I also wonder (and @robrix might know here)—can we use the literal `[d| … ]` quasiquotation style here? It is significantly more readable, though I don't know if there are any technical reasons why we can't.

src/CodeGen/GenerateSyntax.hs
patrickt: Aha. @aymannadeem asked why we didn't use `addTickIfNecessary` there, and I could not come up with a good answer.

src/TreeSitter/Importing.hs
patrickt: I much prefer `Unmarshal` over `Building`.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+instance Building Text.Text where
+  buildNode = do
+    node <- peekNode
+    case node of
+      Just node' -> do
+        bytestring <- ask
+        let start = fromIntegral (nodeStartByte node')
+            end = fromIntegral (nodeEndByte node')
+        pure (decodeUtf8 (slice start end bytestring))
+      _ -> fail "expected a node for Text"
```
patrickt: Can you clarify this error message a little? I wouldn't know what it meant if I hit it.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+
+instance GBuilding f => GBuilding (M1 C c f) where
+  gbuildNode = M1 <$> gbuildNode
+
+-- Possibly for anonymous leaf nodes:
+instance GBuilding U1 where
+  gbuildNode = pure U1
+
+-- For regular leaf nodes
+instance {-# Overlappable #-} GBuilding (M1 S s (K1 c Text.Text)) where
```
patrickt: Why do we need the Overlappable pragmas here? (Also, small stylistic nit: pragmas should be capitalized, e.g. `OVERLAPS` and `OVERLAPPABLE`.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,271 @@
...
+    gbuildProductNode fields
+
+-- For sum datatypes:
+instance (GBuildingSum f, GBuildingSum g, SymbolMatching f, SymbolMatching g) => GBuilding (f :+: g) where
+  gbuildNode = gbuildSumNode @(f :+: g)
+
+-- For product datatypes:
+instance (GBuildingProduct f, GBuildingProduct g) => GBuilding (f :*: g) where
+  gbuildNode = push $ do
+    fields <- getFields
```
patrickt:
```suggestion
    getFields >>= gBuildProductNode @(f :+: g)
```

src/TreeSitter/Importing.hs
robrix: Agreed, tho ultimately this is @aymannadeem’s call.

languages/python/tree-sitter-python.cabal
robrix: Good catch, thank you!

src/CodeGen/GenerateSyntax.hs
robrix: Unfortunately not; AFAIK, `[d|…|]` won’t let us splice in the datatype’s name.

We could maybe simplify a little by using the smart constructors, tho, e.g. `dataD` & friends.

src/TreeSitter/Importing.hs
robrix: It’s cos this instance overlaps unavoidably with the one for unary products.

src/CodeGen/GenerateSyntax.hs
robrix: This might indicate an opportunity for a better name ☝️

src/TreeSitter/Importing.hs
aymannadeem: Because we want the compiler to use `Text.Text` here in the case of regular leaf nodes, but if you look at the unary product `instance {-# Overlappable #-} (Selector s, Building k) => GBuilding (M1 S s (K1 c k))`, we have some polymorphic `k` that we _never_ want as `Text.Text`.

src/CodeGen/GenerateSyntax.hs
aymannadeem: @patrickt I went with the longhand because I couldn't find official TH documentation on the quasi quotation style, so this was easier to debug and find references in Dash for. I agree `[d|...]` is more readable and shorter but if I have to come back to this code, I will rely on Dash more than my memory. 

I should def factor out the duplication though.

src/TreeSitter/Importing.hs
aymannadeem: Documenting this as a learning thing. When I tried to apply this suggestion with visible type application, I got this vague error message with some auto-generated `t1`:

```

src/TreeSitter/Importing.hs:227:37: error:
    • Cannot apply expression of type ‘t1’
      to a visible type argument ‘(f :+: g)’
    • In the second argument of ‘(>>=)’, namely
        ‘gBuildProductNode @(f :+: g)’
      In the second argument of ‘($)’, namely
        ‘getFields >>= gBuildProductNode @(f :+: g)’
      In the expression:
        push $ getFields >>= gBuildProductNode @(f :+: g)
    |
227 |   gbuildNode = push $ getFields >>= gBuildProductNode @(f :+: g)
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cabal: Failed to build tree-sitter-0.1.0.0 (which is required by
tree-sitter-python-0.1.0.1).
```

Removing the type application clarifies it, and it's about the identifier not being in scope: 
```
src/TreeSitter/Importing.hs:227:37: error:
    • Variable not in scope:
        gBuildProductNode :: Map.Map FieldName Node -> m ((:*:) f g a)
    • Perhaps you meant ‘gbuildProductNode’ (line 256)
    |
227 |   gbuildNode = push $ getFields >>= gBuildProductNode
    |                                     ^^^^^^^^^^^^^^^^^
cabal: Failed to build tree-sitter-0.1.0.0 (which is required by
tree-sitter-python-0.1.0.1).
```

src/CodeGen/GenerateSyntax.hs
aymannadeem: I intend to handle this and the refactoring related to consolidating `toName` and `toName'` on a separate branch.

src/CodeGen/GenerateSyntax.hs
aymannadeem: Also going to handle as part of `toName` refactor in a separate PR.

src/CodeGen/GenerateSyntax.hs
aymannadeem: Will handle on separate PR.

src/TreeSitter/Importing.hs
aymannadeem: Will handle in a separate PR!

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,267 @@
--- WIP: Do not depend on this yet!
-{-# LANGUAGE DeriveFunctor, FlexibleContexts, GeneralizedNewtypeDeriving, ScopedTypeVariables, TypeApplications #-}
-module TreeSitter.Importing where
+{-# LANGUAGE DefaultSignatures, FlexibleContexts, FlexibleInstances,
+             PolyKinds, ScopedTypeVariables, TypeApplications, TypeOperators #-}
+module TreeSitter.Importing
```
robrix: A note probably for a future PR: we should rename the module to something like `TreeSitter.Unmarshal` now.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,267 @@
...
+      withCursor (castPtr rootPtr) $ \ cursor ->
+        runM (runFail (runReader cursor (runReader bytestring buildNode)))
 
-data Number = Number
-      deriving (Eq, Ord, Show)
+-- | Unmarshal is the process of iterating over tree-sitter’s parse trees using its tree cursor API and producing Haskell ASTs for the relevant nodes.
+--
+--   Datatypes which can be constructed from tree-sitter parse trees may use the default definition of 'buildNode' providing that they have a suitable 'Generic' instance.
+class Unmarshal a where
+  buildNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m a
```
robrix: Let’s rename `build*` to `unmarshal*` as well 👍 (Whether in this PR or a future one.)

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,267 @@
...
+  deriving (Eq, Ord, Show)
+
+
+-- | Generic construction of ASTs from a 'Map.Map' of named fields.
+--
+--   Product types (specifically, record types) are constructed by looking up the node for each corresponding field name in the map, moving the cursor to it, and then invoking 'buildNode' to construct the value for that field. Leaf types are constructed as a special case of product types.
+--
+--   Sum types are constructed by attempting to build each constructor nondeterministically. This should instead use the current node’s symbol to select the corresponding constructor deterministically.
+class GUnmarshal f where
+  gbuildNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m (f a)
```
robrix: As above, `gbuild*` should probably be `gunmarshal*`.

src/TreeSitter/Importing.hs
```diff
@@ -1,135 +1,267 @@
...
-instance Applicative m => Applicative (MaybeC m) where
-  pure a = MaybeC (pure (Just a))
-  liftA2 f (MaybeC a) (MaybeC b) = MaybeC $ liftA2 (liftA2 f) a b
+instance (GUnmarshalSum f, GUnmarshalSum g, SymbolMatching f, SymbolMatching g) => GUnmarshalSum (f :+: g) where
+  gbuildSumNode = do
+    currentNode <- peekNode
+    (lhsSymbolMatch, rhsSymbolMatch, currentNode) <- case currentNode of
+      Just node -> pure (symbolMatch (Proxy @f) node, symbolMatch (Proxy @g) node, node)
+      Nothing -> fail "expected a node; got none"
+    if lhsSymbolMatch -- FIXME: report error
```
robrix: This `FIXME` is obsolete.

robrix:
gif of Mabel Pines from Gravity Falls saying “the lamby dance!” and looking all happy

PR #153, opened 2019-06-28
Refactor toName/toName' and friends

aymannadeem:
There's a lot of duplication in the code used to convert the `snake_case` input strings into `Name` types that are commensurable with further TH munging we need to do to generate Haskell syntaxes. This is a mess I made, so I'm in the process of cleaning it up.


aymannadeem:
I'm done!

Issue #99, opened 2019-05-17
Comments aren’t represented in AST datatypes


robrix:
AST datatypes don’t currently have any way to represent comments.

robrix:
Some discussion today w/ @aymannadeem & @maxbrunsfeld led us to conclude that adding leading/trailing extras fields to product & leaf types should be a good way to represent this, tho we also have to deal with populating said fields which could be a bit annoying.

We could also maybe deal with it using higher-kinded data, cf #143 for some more details on that.

robrix:
If we didn’t care about leading/trailing so much, then it would be _really_ easy to add comments to products & leaves by unmarshalling anything for which both `ts_node_is_named` & `ts_node_is_extra` return a truthy value into a field named `extras`. I think the only problem would be figuring out what the type should be, but we might be able to figure that out from the `node-types.json` data already.

robrix:
(Note that this would probably attribute all comments to their immediate parent node, which would make it much harder to relate doc comments to the nodes they’re documenting.)

robrix:
Depending on where comments get attributed (cc @maxbrunsfeld), this could also be a good use case for HKD as discussed in #263 if we used a functor à la:

```haskell
data Commented a = Commented
  { leading  :: [Comment]
  , body     :: a
  , trailing :: [Comment]
  }
```

PR #203, opened 2019-09-16
Skip extras in precise ASTs (for now)

### LINKED ISSUE #204: Comments are mistakenly being unmarshalled into extraChildren fields
We don’t currently have support for extras like comments (cf #99), but in grammars that have named comment rules, the comments get added into the extra children field, which is causing Python modules to fail to unmarshal since `Either CompoundStatement SimpleStatement` can’t represent `Comment`.

robrix:
This PR fixes a bug where named nodes occurring within extras were being added to the extra children field, which was definitely a bug altho as @maxbrunsfeld notes is kinda confusing in terms of nomenclature.

Since we haven’t yet got a fix for #99, we just drop them for now.

Fixes #204.


tree-sitter-python/test/Main.hs
```diff
@@ -31,6 +31,7 @@ function = Py.ExpressionStatementSimpleStatement (Py.ExpressionStatement () [Lef
 prop_simpleExamples :: Property
 prop_simpleExamples = property $ do
   "" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [] }
+  "# bah" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [] }
```
robrix: This test fails on `master`.

tree-sitter/src/TreeSitter/Node.hs
```diff
@@ -24,6 +24,7 @@ data Node = Node
   , nodeEndByte    :: !Word32
   , nodeChildCount :: !Word32
   , nodeIsNamed    :: !CBool
+  , nodeIsExtra    :: !CBool
```
robrix: Thank you @maxbrunsfeld! 💖

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -221,7 +221,8 @@ getFields = go Map.empty -- >>= \fields -> liftIO (print (Map.keys fields)) >> p
               keepGoing <- step
               let fs' = case fieldName of
                     Just fieldName' -> Map.insertWith (flip (++)) fieldName' [node'] fs
-                    _ -> if nodeIsNamed node' /= 0
+                    -- NB: We currently skip “extra” nodes (i.e. ones occurring in the @extras@ rule), pending a fix to https://github.com/tree-sitter/haskell-tree-sitter/issues/99
+                    _ -> if nodeIsNamed node' /= 0 && nodeIsExtra node' == 0
                       then Map.insertWith (flip (++)) (FieldName "extraChildren") [node'] fs
                       else fs
```
robrix: The addition of the `nodeIsExtra` test means that we just silently drop such fields. It won’t be a big change to nest the conditionals and add extras into a dedicated `extras` field instead, once we’re fixing #99. @maxbrunsfeld made the (v. good) point that having both `extras` and `extraChildren` fields is gonna be super confusing, so maybe we should also rename the latter to `otherChildren` at that point.

tree-sitter/test/Spec.hs
```diff
@@ -30,8 +30,8 @@ prop_Node_sizeOf = property $
   sizeOf (undefined :: Node) === fromIntegral sizeof_node
 
 prop_Node_roundtrips = property $ do
-  peeked <- liftIO (with (Node (TSNode 1 2 3 4 nullPtr nullPtr) nullPtr 1 (TSPoint 2 3) (TSPoint 4 5) 6 7 8 9) peek)
-  peeked ===              Node (TSNode 1 2 3 4 nullPtr nullPtr) nullPtr 1 (TSPoint 2 3) (TSPoint 4 5) 6 7 8 9
+  peeked <- liftIO (with (Node (TSNode 1 2 3 4 nullPtr nullPtr) nullPtr 1 (TSPoint 2 3) (TSPoint 4 5) 6 7 8 9 10) peek)
+  peeked ===              Node (TSNode 1 2 3 4 nullPtr nullPtr) nullPtr 1 (TSPoint 2 3) (TSPoint 4 5) 6 7 8 9 10
```
robrix: look at all the fields I give

look at them, Anakin

tree-sitter-go/tree-sitter-go.cabal
```diff
@@ -14,16 +14,16 @@ build-type:          Simple
...
                      , tree-sitter-go-internal
   default-language:    Haskell2010
 
 library tree-sitter-go-internal
   exposed-modules:     TreeSitter.Go.Internal
   hs-source-dirs:      internal
-  build-depends:       base                 >= 4.12 && < 5
-                     , tree-sitter >= 0.1 && < 0.3
+  build-depends:       base
+                     , tree-sitter
```
robrix: Internal libraries can’t be used outside of the package they’re defined in, and any constraints they impose are effectively unioned with the public library, so I’ve just deleted them all and allowed the public library components to constrain things instead.

tree-sitter-go/tree-sitter-go.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:       2.2
 name:                tree-sitter-go
-version:             0.1.0.0
+version:             0.2.0.0
```
robrix: This is a major bump because it requires a major bump, I guess?

CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
...
 ### 0.1.0
 
 * Support for tree-sitter-go
 
 ### 0.0.0.1
 * Support for tree-sitter-javascript
 * Support for tree-sitter-c
-* Preliminary support for tree-sitter-ruby
\ No newline at end of file
+* Preliminary support for tree-sitter-ruby
```
robrix: Trailing newline, apparently.

PR #212, opened 2019-09-19
Bump tree-sitter-typescript

tclem:
Pick up fix for state merging that was causing some parse failures. https://github.com/tree-sitter/tree-sitter-typescript/commit/2573038b6c77840618df17064800c914234207cb.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -2,4 +2,4 @@
 
 git submodule sync && git submodule update --init --recursive
 
-cabal update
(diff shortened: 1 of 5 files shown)


tclem:
Published:

- [x] https://hackage.haskell.org/package/tree-sitter-typescript-0.2.1.0
- [ ] https://hackage.haskell.org/package/tree-sitter-tsx-0.2.1.0

Issue #210, opened 2019-09-19
Define type synonyms in lieu of sum datatypes


robrix:
`TreeSitter.Python.AST.Module` has a single field of type `[Either CompoundStatement SimpleStatement]`. Each of these are themselves sum datatypes:

```haskell
data SimpleStatement a
  = AssertStatementSimpleStatement (AssertStatement a)
  | BreakStatementSimpleStatement (BreakStatement a)
  | ContinueStatementSimpleStatement (ContinueStatement a)
  | DeleteStatementSimpleStatement (DeleteStatement a)
  | ExecStatementSimpleStatement (ExecStatement a)
  | ExpressionStatementSimpleStatement (ExpressionStatement a)
  | FutureImportStatementSimpleStatement (FutureImportStatement a)
  | GlobalStatementSimpleStatement (GlobalStatement a)
  | ImportFromStatementSimpleStatement (ImportFromStatement a)
  | ImportStatementSimpleStatement (ImportStatement a)
  | NonlocalStatementSimpleStatement (NonlocalStatement a)
  | PassStatementSimpleStatement (PassStatement a)
  | PrintStatementSimpleStatement (PrintStatement a)
  | RaiseStatementSimpleStatement (RaiseStatement a)
  | ReturnStatementSimpleStatement (ReturnStatement a)
```

Right now we duplicate a bunch of the logic used on the one hand for named sum datatypes and on the other hand for anonymous sums like the `Either` mentioned above. Furthermore, sum datatypes have to have lengthy, duplicative constructor names so that they don’t conflict with the names of the constructors they contain.

We could simplify things significantly by just not defining sum datatypes at all, and instead defining type synonyms for an anonymous sum, e.g.:

```haskell
type SimpleStatement a
  = Either (AssertStatement a)
  ( Either (BreakStatement a)
  ( Either (ContinueStatement a)
  ( Either (DeleteStatement a)
  ( Either (ExecStatement a)
  ( Either (ExpressionStatement a)
  ( Either (FutureImportStatement a)
  ( Either (GlobalStatement a)
  ( Either (ImportFromStatement a)
  ( Either (ImportStatement a)
  ( Either (NonlocalStatement a)
  ( Either (PassStatement a)
  ( Either (PrintStatement a)
  ( Either (RaiseStatement a)
  ( Either (ReturnStatement a)))))))))))))))
```

[...]

PR #221, opened 2019-09-25
Faster unmarshalling iteration

### LINKED ISSUE #220: Validate Unmarshaling sooner/from tree-sitter
`Unmarshal` lives in `tree-sitter`, but we don't really get to test it until we load `tree-sitter-python`. This can delay useful feedback. To make sure we see the errors closer to the module where they live, @robrix proposed hard-coding syntax types that emulate what our TH-generated types would look like, ie.:

```
data Stmt a
  = IfStmt (If a)
  | BlockStmt (Block a)
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)

data Expr a
  = LitExpr (Lit a)
  | PlusExpr (Plus a)
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)

data Lit a = Lit { ann :: a, text :: Text.Text }
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)
data Plus a = Plus { ann :: a, lhs :: Expr a, rhs :: Expr a }
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)

data Block a = Block { ann :: a, body :: [Stmt a] }
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)

data If a = If { ann :: a, condition :: Expr a, consequence :: Stmt a, alternative :: Stmt a }
  deriving (Eq, Ord, Show, Foldable, Functor, Traversable, Generic1)
```

This would allow us to test/validate the unmarshal code is working without having to load language-specific projects where the unmarshaling is actually happening right now.

robrix:
This PR defines some datatypes and derives `Unmarshal` instances for them so that we can get immediate feedback on whether we’re actually able to derive the behaviours we want for the kinds of datatypes we generate with TH.

**NB:** the thing being optimized here is our workflow when iterating on `Unmarshal`, rather than e.g. the runtime performance of unmarshalling.

- [x] Fixes #220.


tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -0,0 +1,55 @@
+{-# LANGUAGE DeriveAnyClass, DeriveGeneric, DuplicateRecordFields, TypeOperators #-}
+module TreeSitter.Unmarshal.Examples () where
```
robrix: Adding these examples to the library means that we get instantaneous feedback when changes to `Unmarshal` affect which instances we can derive.

tree-sitter/tree-sitter.cabal
```diff
@@ -47,6 +47,7 @@ library
                      , TreeSitter.GenerateSyntax
                      , TreeSitter.Deserialize
                      , TreeSitter.Cursor
+  other-modules:       TreeSitter.Unmarshal.Examples
```
robrix: We don’t want to actually expose these example datatypes to downstream consumers, so we include them in `other-modules` instead of `exposed-modules`.

tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -0,0 +1,62 @@
...
+-- | An example of a sum-of-products datatype.
+data Expr a
+  = IfExpr (If a)
+  | BlockExpr (Block a)
+  | LitExpr (Lit a)
+  | BinExpr (Bin a)
+  deriving (Generic, Unmarshal)
+
+-- | Product with multiple fields.
+data If a = If { ann :: a, condition :: Expr a, consequence :: Expr a, alternative :: Maybe (Expr a) }
```
robrix: These datatypes are intended to be representative, but not exhaustive. We employ sums, products, leaves, and anonymous leaves; as well as `Maybe`, `[]`, and `:+:`; but not `NonEmpty`.

tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -0,0 +1,62 @@
...
+import qualified Data.Text as Text
+import GHC.Generics ((:+:), Generic)
+import TreeSitter.Unmarshal
+
+-- | An example of a sum-of-products datatype.
+data Expr a
+  = IfExpr (If a)
+  | BlockExpr (Block a)
+  | LitExpr (Lit a)
+  | BinExpr (Bin a)
```
robrix: (When we implement #210, we’ll want to come back and replace this with a type synonym instead.)

PR #222, opened 2019-09-25
Unmarshal at * -> *

### LINKED ISSUE #219: symbolMatchingInstanceForSums should be kind `* -> *`
Noting this so I don't forget. As @robrix pointed out, `symbolMatchingInstanceForSums` currently generates kind `*` but we (potentially?) want `* -> *`,  just like we did for anonymous sums.

robrix:
This changes `Unmarshal` to operate on type constructors of kind * -> * instead of types of kind *, allowing us a great deal more latitude in separating out the various roles types play in our ASTs. This was made possible by @aymannadeem’s work in #214, and results in a substantial simplification to the implementation, which just goes to show what you can accomplish with parametricity.

- [x] Depends on #221.
- [x] Fixes #219.

### DIFF
--- a/tree-sitter/src/TreeSitter/Unmarshal.hs
+++ b/tree-sitter/src/TreeSitter/Unmarshal.hs
@@ -38,7 +38,7 @@ import           Source.Span
 import           Source.Range
 import           Data.Proxy
 import           Prelude hiding (fail)
-import           Data.Maybe (fromMaybe, maybeToList)
+import           Data.Maybe (fromMaybe)
 import           Data.List.NonEmpty (NonEmpty (..))
 
 -- Parse source code and produce AST
@@ -49,7 +49,7 @@ parseByteString language bytestring = withParser language $ \ parser -> withPars
   else
     withRootNode treePtr $ \ rootPtr ->
       withCursor (castPtr rootPtr) $ \ cursor ->
-        runM (runFail (runReader cursor (runReader bytestring (peekNode >>= unmarshalNodes . maybeToList))))
+        runM (runFail (runReader cursor (runReader bytestring (peekNode >>= maybe (fail "expected a root node") unmarshalNode))))
 
 -- | Unmarshal is the process of iterating over tree-sitter’s parse trees using its tree cursor API and producing Haskell ASTs for the relevant nodes.
 --
--- a/tree-sitter/src/TreeSitter/Unmarshal.hs
+++ b/tree-sitter/src/TreeSitter/Unmarshal.hs
@@ -333,13 +333,38 @@ instance (GUnmarshalSum f, GUnmarshalSum g, SymbolMatching f, SymbolMatching g) 
 
 -- | Generically unmarshal products
 class GUnmarshalProduct f where
-  gunmarshalProductNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => Map.Map FieldName [Node] -> m (f a)
+  gunmarshalProductNode
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       , UnmarshalAnn a
+       )
+    => Node
+    -> Map.Map FieldName [Node]
+    -> m (f a)
 
 -- Product structure
 instance (GUnmarshalProduct f, GUnmarshalProduct g) => GUnmarshalProduct (f :*: g) where
-  gunmarshalProductNode fields = (:*:) <$> gunmarshalProductNode @f fields <*> gunmarshalProductNode @g fields
+  gunmarshalProductNode node fields = (:*:)
+    <$> gunmarshalProductNode @f node fields
+    <*> gunmarshalProductNode @g node fields
 
 -- Contents of product types (ie., the leaves of the product tree)
-instance (Unmarshal k, Selector c) => GUnmarshalProduct (M1 S c (K1 i k)) where
-  gunmarshalProductNode fields =
-   M1 . K1 <$> unmarshalNodes (fromMaybe [] (Map.lookup (FieldName (selName @c undefined)) fields))
+instance UnmarshalAnn k => GUnmarshalProduct (M1 S c (K1 i k)) where
+  gunmarshalProductNode node _ = M1 . K1 <$> unmarshalAnn node
+
+instance GUnmarshalProduct (M1 S c Par1) where
+  gunmarshalProductNode node _ = M1 . Par1 <$> unmarshalAnn node
+
+instance (UnmarshalField f, Unmarshal g, Selector c) => GUnmarshalProduct (M1 S c (f :.: g)) where
+  gunmarshalProductNode _ fields =
+    M1 . Comp1 <$> unmarshalField (fromMaybe [] (Map.lookup (FieldName (selName @c undefined)) fields))
+
+instance (Unmarshal t, Selector c) => GUnmarshalProduct (M1 S c (Rec1 t)) where
+  gunmarshalProductNode _ fields =
+    case fromMaybe [] (Map.lookup (FieldName (selName @c undefined)) fields) of
+      []  -> fail "expected a node but didn't get one"
+      [x] -> M1 . Rec1 <$> unmarshalNode x
+      _   -> fail "expected a node but got multiple"
--- a/tree-sitter/src/TreeSitter/Unmarshal.hs
+++ b/tree-sitter/src/TreeSitter/Unmarshal.hs
@@ -284,7 +284,16 @@ newtype FieldName = FieldName { getFieldName :: String }
 --
 --   Sum types are constructed by attempting to unmarshal each constructor nondeterministically. This should instead use the current node’s symbol to select the corresponding constructor deterministically.
 class GUnmarshal f where
-  gunmarshalNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => Node -> m (f a)
+  gunmarshalNode
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       , UnmarshalAnn a
+       )
+    => Node
+    -> m (f a)
 
 instance GUnmarshal f => GUnmarshal (M1 D c f) where
   gunmarshalNode node = M1 <$> gunmarshalNode node
@@ -298,8 +307,12 @@ instance GUnmarshal U1 where
 
 
 -- For unary products:
-instance (Selector s, Unmarshal k) => GUnmarshal (M1 S s (K1 c k)) where
-  gunmarshalNode _ = push getFields >>= gunmarshalProductNode . fromMaybe Map.empty
+instance (Selector s, UnmarshalAnn k) => GUnmarshal (M1 S s (K1 c k)) where
+  gunmarshalNode node = push getFields >>= gunmarshalProductNode node . fromMaybe Map.empty
+
+-- For anonymous leaf nodes
+instance GUnmarshal (M1 S s Par1) where
+  gunmarshalNode = fmap (M1 . Par1) <$> unmarshalAnn
 
 -- For sum datatypes:
 instance (GUnmarshalSum f, GUnmarshalSum g, SymbolMatching f, SymbolMatching g) => GUnmarshal (f :+: g) where
@@ -307,7 +320,7 @@ instance (GUnmarshalSum f, GUnmarshalSum g, SymbolMatching f, SymbolMatching g) 
 
 -- For product datatypes:
 instance (GUnmarshalProduct f, GUnmarshalProduct g) => GUnmarshal (f :*: g) where
-  gunmarshalNode _ = push getFields >>= gunmarshalProductNode @(f :*: g) . fromMaybe Map.empty
+  gunmarshalNode node = push getFields >>= gunmarshalProductNode @(f :*: g) node . fromMaybe Map.empty
 
 class GUnmarshalSum f where
   gunmarshalSumNode
--- a/tree-sitter/src/TreeSitter/Unmarshal.hs
+++ b/tree-sitter/src/TreeSitter/Unmarshal.hs
@@ -4,6 +4,8 @@ module TreeSitter.Unmarshal
 ( parseByteString
 , FieldName(..)
 , Unmarshal(..)
+, UnmarshalAnn(..)
+, UnmarshalField(..)
 , SymbolMatching(..)
 , step
 , push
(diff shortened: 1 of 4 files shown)


tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+  unmarshalNode x = do
     goto (nodeTSNode x)
-    to <$> gunmarshalNode x
-  unmarshalNodes [] = fail "expected a node but didn't get one"
-  unmarshalNodes _ = fail "expected a node but got multiple"
+    to1 <$> gunmarshalNode x
 
-instance Unmarshal () where
-  unmarshalNodes _ = pure ()
+instance (Unmarshal f, Unmarshal g, SymbolMatching f, SymbolMatching g) => Unmarshal (f :+: g) where
```
robrix: This is the same instance as we had before, just at `* -> *` instead.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
-  unmarshalNodes [x] = do
+--   Datatypes which can be constructed from tree-sitter parse trees may use the default definition of 'unmarshalNode' providing that they have a suitable 'Generic1' instance.
+class Unmarshal t where
+  unmarshalNode
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       , UnmarshalAnn a
```
robrix: Because `t` is `* -> *`, we require its parameter to have an `UnmarshalAnn` instance.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+--   Leaf nodes have 'Text.Text' fields, and leaves, anonymous leaves, and products all have parametric annotation fields. All of these fields are unmarshalled using the metadata of the node, e.g. its start/end bytes, without reference to any child nodes it may contain.
+class UnmarshalAnn a where
+  unmarshalAnn
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       )
+    => Node
```
robrix: `UnmarshalAnn` is also singular.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+    -> m a
+
+instance UnmarshalAnn () where
+  unmarshalAnn _ = pure ()
+
+instance UnmarshalAnn Text.Text where
+  unmarshalAnn node = do
```
robrix: Because `unmarshalAnn` takes the node instead of having to `peekNode`, it’s much simpler and doesn’t have to deal with the possibility of failure. Happily, this also entails less crossing of the FFI!

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+    range <- unmarshalAnn node
```
robrix: We take advantage of the instance for `Range` here.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+  unmarshalField
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       , Unmarshal f
+       , UnmarshalAnn a
+       )
+    => [Node]
```
robrix: `UnmarshalField` deals with “modifier” types composed onto AST types: fields wrapped in `Maybe`, `[]`, or `NonEmpty`. These are the only instances which have to deal with zero-or-more nodes.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
-class SymbolMatching a where
+class SymbolMatching (a :: * -> *) where
   symbolMatch :: Proxy a -> Node -> Bool
 
   -- | Provide error message describing the node symbol vs. the symbols this can match
   showFailure :: Proxy a -> Node -> String
 
-instance SymbolMatching a => SymbolMatching (Maybe a) where
-  symbolMatch _ = symbolMatch (Proxy @a)
-  showFailure _ = showFailure (Proxy @a)
```
robrix: A bunch of these `SymbolMatching` instances are completely unnecessary now: we only use `SymbolMatching` for sums, and these modifier types only occur in products.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -38,131 +40,161 @@ import           Source.Span
...
+
+instance UnmarshalField Maybe where
+  unmarshalField []  = pure Nothing
+  unmarshalField [x] = Just <$> unmarshalNode x
+  unmarshalField _   = fail "expected a node of type ((f :+: g) a) but got multiple"
+
+instance UnmarshalField [] where
+  unmarshalField (x:xs) = do
+    head' <- unmarshalNode x
+    tail' <- unmarshalField xs
```
robrix: The control flow here is a lot more obvious than it was before: we unmarshal each node in the field to an element in the list.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -186,19 +218,15 @@ goto node = do
   cursor <- ask
   liftIO (with node (ts_tree_cursor_reset_p cursor))
 
--- | Return the 'Node' that the cursor is pointing at (if any), or 'Nothing' otherwise.
-peekNode :: (Carrier sig m, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m (Maybe Node)
+-- | Return the 'Node' that the cursor is pointing at.
+peekNode :: (Carrier sig m, Member (Reader (Ptr Cursor)) sig, MonadIO m) => m Node
```
robrix: When originally written, we had used API which walked the tree using the next sibling API, which indicates that it’s reached the end by returning a node for which `ts_node_is_null` returns true. By contrast, the cursor API, which we now rely on exclusively, is guarded: API to change the cursor returns a boolean value indicating whether the change occurred. Thus, the cursor API won’t return null nodes, and we don’t have to return in `Maybe`, enabling further tidying.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -210,28 +238,32 @@ peekFieldName = do
...
+
+lookupField :: FieldName -> Fields -> [Node]
+lookupField k = fromMaybe [] . Map.lookup k
+
 
 -- | Return a 'ByteString' that contains a slice of the given 'ByteString'.
-slice :: Int -> Int -> ByteString -> ByteString
-slice start end = take . drop
+slice :: Range -> ByteString -> ByteString
+slice (Range start end) = take . drop
```
robrix: We should eventually change this all to operate on `Source` instead.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
-      then L1 <$> gunmarshalSumNode @f node
-      else if rhsSymbolMatch
-        then R1 <$> gunmarshalSumNode @g node
-        else fail $ showFailure (Proxy @f) node `sep` showFailure (Proxy @g) node
+    if lhsSymbolMatch then
+      L1 <$> gunmarshalNode @f node
+    else if rhsSymbolMatch then
+      R1 <$> gunmarshalNode @g node
+    else
+      fail $ showFailure (Proxy @f) node `sep` showFailure (Proxy @g) node
```
robrix: I didn’t actually change this logic, just reformatted the `if` statement for alignment.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
+  gunmarshalNode node = Par1 <$> unmarshalAnn node
+
+instance (Unmarshal t, SymbolMatching t) => GUnmarshal (Rec1 t) where
+  gunmarshalNode node = Rec1 <$> unmarshalNode node
 
 -- For product datatypes:
 instance (GUnmarshalProduct f, GUnmarshalProduct g) => GUnmarshal (f :*: g) where
-  gunmarshalNode _ = push getFields >>= gunmarshalProductNode @(f :*: g) . fromMaybe Map.empty
-
-class GUnmarshalSum f where
```
robrix: Happily, `GUnmarshalSum` is no longer necessary _at all_. We just use `GUnmarshal` to handle sums.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
+    else if rhsSymbolMatch then
+      R1 <$> gunmarshalNode @g node
+    else
+      fail $ showFailure (Proxy @f) node `sep` showFailure (Proxy @g) node
 
 
 -- | Generically unmarshal products
 class GUnmarshalProduct f where
-  gunmarshalProductNode :: (MonadFail m, Carrier sig m, Member (Reader ByteString) sig, Member (Reader (Ptr Cursor)) sig, MonadIO m) => Map.Map FieldName [Node] -> m (f a)
+  gunmarshalProductNode
```
robrix: We still retain `GUnmarshalProduct` to deal with some of the quirks of products, e.g. field names being meaningful, and to avoid attempting to collect the fields for datatypes which don’t need them (where possible).

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
+    :: ( Carrier sig m
+       , Member (Reader ByteString) sig
+       , Member (Reader (Ptr Cursor)) sig
+       , MonadFail m
+       , MonadIO m
+       , UnmarshalAnn a
+       )
+    => Node
```
robrix: `GUnmarshalProduct` now takes the `Node` as well as the `Fields`, allowing for more direct handling of `Text` & annotation fields.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
-instance (Unmarshal k, Selector c) => GUnmarshalProduct (M1 S c (K1 i k)) where
-  gunmarshalProductNode fields =
-   M1 . K1 <$> unmarshalNodes (fromMaybe [] (Map.lookup (FieldName (selName @c undefined)) fields))
+instance UnmarshalAnn k => GUnmarshalProduct (M1 S c (K1 i k)) where
+  gunmarshalProductNode node _ = M1 . K1 <$> unmarshalAnn node
+
+instance GUnmarshalProduct (M1 S c Par1) where
+  gunmarshalProductNode node _ = M1 . Par1 <$> unmarshalAnn node
+
+instance (UnmarshalField f, Unmarshal g, Selector c) => GUnmarshalProduct (M1 S c (f :.: g)) where
```
robrix: Optional/multiple fields are now exclusively handled via this instance. This means that we could technically eliminate `UnmarshalField` altogether, replacing it with multiple specializations of this instance. However, it would be a little more challenging to share code between them, and it would needlessly constrain users who might want to define their own AST datatypes instead of deriving them from the grammar.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -260,48 +298,67 @@ instance GUnmarshal U1 where
...
+  gunmarshalProductNode _ fields =
+    M1 . Comp1 <$> unmarshalField (lookupField (FieldName (selName @c undefined)) fields)
+
+instance (Unmarshal t, Selector c) => GUnmarshalProduct (M1 S c (Rec1 t)) where
+  gunmarshalProductNode _ fields =
+    case lookupField (FieldName (selName @c undefined)) fields of
+      []  -> fail "expected a node but didn't get one"
+      [x] -> M1 . Rec1 <$> unmarshalNode x
+      _   -> fail "expected a node but got multiple"
```
robrix: Because AST fields are now represented with `Rec1`, we know a priori that `t` only takes one node, and thus can handle the error cases immediately instead of pushing them into every `Unmarshal` instance.

tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -1,62 +1,93 @@
...
+  deriving (Generic1, Unmarshal)
 
-instance SymbolMatching (Lit a) where
+instance SymbolMatching Var where
+  symbolMatch _ _ = False
+  showFailure _ _ = ""
+
+-- | Custom leaf node.
+data Lit a = Lit { ann :: a, lit :: IntegerLit }
+  deriving (Generic1, Unmarshal)
```
robrix: I wanted to illustrate something cool the `Unmarshal` machinery can do: it can accommodate datatypes which a) our TH code doesn’t generate, and b) do something more specialized. See the `UnmarshalAnn` instance for `IntegerLit`, below.

tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -1,62 +1,93 @@
...
-  deriving (Generic, Unmarshal)
+  deriving (Generic1, Unmarshal)
 
-instance SymbolMatching (AnonTimes a) where
+instance SymbolMatching AnonTimes where
   symbolMatch _ _ = False
   showFailure _ _ = ""
+
+
+newtype IntegerLit = IntegerLit Integer
```
robrix: `IntegerLit` holds an `Integer`, but we don’t know how to unmarshal those!

tree-sitter/src/TreeSitter/Unmarshal/Examples.hs
```diff
@@ -1,62 +1,93 @@
...
+  unmarshalAnn node = do
+    Range start end <- unmarshalAnn node
+    bytestring <- ask
+    let drop = B.drop start
+        take = B.take (end - start)
+        slice = take . drop
+        str = Text.unpack (Text.decodeUtf8 (slice bytestring))
+    case readDec str of
+      (i, _):_ -> pure (IntegerLit i)
+      _        -> fail ("could not parse '" <> str <> "'")
```
robrix: So, we tell it how: get the range, slice the source down to a string, and use `readDec` to produce an `Integer` _with total error-checking of malformed inputs_.

We’re unlikely to ever have the TH code produce ASTs with `Integer` fields—parsing the contents of these strings is language-specific, after all—but we could imagine having the TH code allow us to specify some datatypes that we want to define by hand instead. This gives us a way to do that which still usefully integrates with `Unmarshal`, meaning that you can derive _most_ of your data, specialize only small parts of it, only do this parsing once, right up front, and still parse and unmarshal the whole thing to Haskell data.

PR #225, opened 2019-09-27
type synonyms

### LINKED ISSUE #210: Define type synonyms in lieu of sum datatypes

aymannadeem:
Fixes #210 

Pardon the verbose output I'm pasting in, but this change results in generating the following type synonyms instead of creating sum datatypes with ugly strings appended:

```
type TreeSitter.Python.AST.SimpleStatement =
  TreeSitter.Python.AST.AssertStatement
  GHC.Generics.:+: (TreeSitter.Python.AST.BreakStatement
                    GHC.Generics.:+: (TreeSitter.Python.AST.ContinueStatement
                                      GHC.Generics.:+: (TreeSitter.Python.AST.DeleteStatement
                                                        GHC.Generics.:+: (TreeSitter.Python.AST.ExecStatement
                                                                          GHC.Generics.:+: (TreeSitter.Python.AST.ExpressionStatement
                                                                                            GHC.Generics.:+: (TreeSitter.Python.AST.FutureImportStatement
                                                                                                              GHC.Generics.:+: (TreeSitter.Python.AST.GlobalStatement
                                                                                                                                GHC.Generics.:+: (TreeSitter.Python.AST.ImportFromStatement
                                                                                                                                                  GHC.Generics.:+: (TreeSitter.Python.AST.ImportStatement
                                                                                                                                                                    GHC.Generics.:+: (TreeSitter.Python.AST.NonlocalStatement
                                                                                                                                                                                      GHC.Generics.:+: (TreeSitter.Python.AST.PassStatement
                                                                                                                                                                                                        GHC.Generics.:+: (TreeSitter.Python.AST.PrintStatement
                                                                                                                                                                                                                          GHC.Generics.:+: (TreeSitter.Python.AST.RaiseStatement
                                                                                                                                                                                                                                            GHC.Generics.:+: TreeSitter.Python.AST.ReturnStatement)))))))))))))
  :: * -> *
  	-- Defined at TreeSitter/Python/AST.hs:8:1
```

Edit: as per @robrix's [comment below](https://github.com/tree-sitter/haskell-tree-sitter/pull/225#issuecomment-536129502), we decided to move forward with `newtype` wrappers instead.


tree-sitter/src/TreeSitter/Deserialize.hs
```diff
@@ -28,7 +28,7 @@ data Datatype
   = SumType
   { datatypeName       :: DatatypeName
   , datatypeNameStatus :: Named
-  , datatypeSubtypes   :: [Type]
+  , datatypeSubtypes   :: NonEmpty Type
```
aymannadeem: `fieldTypesToNestedSum` takes a `NonEmpty` list, and we want to make sure that the type synonym isn't empty.

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -51,10 +50,9 @@ syntaxDatatype :: Ptr TS.Language -> Datatype -> Q [Dec]
 syntaxDatatype language datatype = do
   typeParameterName <- newName "a"
   case datatype of
-    SumType (DatatypeName datatypeName) _ subtypes -> do
-      cons <- traverse (constructorForSumChoice datatypeName typeParameterName) subtypes
-      result <- symbolMatchingInstanceForSums name subtypes
```
aymannadeem: It's a type synonym now, so no more instances.

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -128,8 +111,8 @@ ctorForTypes named constructorName types = recC (toName named constructorName) r
 
 
 -- | Convert field types to Q types
-fieldTypesToNestedEither :: NonEmpty TreeSitter.Deserialize.Type -> Name -> Q TH.Type
-fieldTypesToNestedEither xs typeParameterName = foldr1 combine (fmap convertToQType xs) `appT` varT typeParameterName
+fieldTypesToNestedSum :: NonEmpty TreeSitter.Deserialize.Type -> Q TH.Type
+fieldTypesToNestedSum xs = foldr1 combine (fmap convertToQType xs)
```
aymannadeem: Prior to changing the type signature/what args this function accepts, I wasn't understanding the error message I was seeing until @robrix clarified the behavior was related to type synonyms, so writing down the reasoning here. 

Suppose you have the following:
```
type Foo = (Maybe :+: Maybe) a
type Bar a = (Foo :+: Maybe) a
```

This won't compile, because type synonyms can't be partially applied.

By contrast, this compiles:
```
type Foo = (Maybe :+: Maybe)

type Bar a = (Foo :+: Maybe) a
```

In the case we need to construct a type with a `typeParameterName`, as we do in the call to `fieldTypesToNestedSums` within `ctorForProductType`, I just use `appT` to create that there.

robrix:
So, after some experimentation tonight, it turns out that using type synonyms for sum fields ends up resulting in _extremely_ slow compiles—I think deriving our various instances for very large `:+:` trees ends up being slower, but I’m not 100% sure yet.

Given that, an intermediate solution might be to generate `newtype`s _wrapping_ `:+:` for sum data types—we’ll still have to derive instances over the `:+:` chains, but only once (for the `newtype`) instead of repeatedly (every time the sum type occurs in a field).

I’m also really curious whether it’s the arity of the trees or the depth that’s the primary factor—i.e. if we generated balanced binary trees instead of right-nested chains, would we be able to derive instances more efficiently? We might want to experiment a bit to see.

tree-sitter-java/TreeSitter/Java/AST.hs
```diff
@@ -1,4 +1,13 @@
-{-# LANGUAGE DataKinds, DeriveAnyClass, DeriveGeneric, DeriveTraversable, DuplicateRecordFields, TemplateHaskell, TypeApplications #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DerivingStrategies #-}
```
robrix: I really love that you’re standardizing this 👍 Would love to hear what motivates your selection of this style when you have the time, obvs. not urgent tho.

tree-sitter-python/test/Main.hs
```diff
@@ -19,10 +19,11 @@ s `shouldParseInto` t = do
...
-plusOne = Py.ExpressionStatementSimpleStatement (Py.ExpressionStatement () [L1 (Py.PrimaryExpressionExpression (Py.UnaryOperatorPrimaryExpression (Py.UnaryOperator () (L1 (Token ())) (Py.IntegerPrimaryExpression (Py.Integer () "1")))))])
-function = Py.ExpressionStatementSimpleStatement (Py.ExpressionStatement () [L1 (Py.PrimaryExpressionExpression (Py.IdentifierPrimaryExpression (Py.Identifier () "expensive")))])
+pass = Py.SimpleStatement (R1 (R1 (L1 (L1 (Py.PassStatement () "pass")))))
+one = Py.SimpleStatement (L1 (R1 (R1 (L1 (Py.ExpressionStatement () [L1 (L1 (Py.Expression (L1 (L1 (L1 (Py.PrimaryExpression (R1 (L1 (L1 (L1 (Py.Integer () "1")))))))))))])))))
+plusOne = Py.SimpleStatement (L1 (R1 (R1 (L1 (Py.ExpressionStatement () [L1 (L1 (Py.Expression (L1 (L1 (L1 (Py.PrimaryExpression (R1 (R1 (R1 (R1 (R1 (Py.UnaryOperator () (L1 (Token ())) (Py.PrimaryExpression (R1 (L1 (L1 (L1 (Py.Integer () "1")))))) ))))))))))))])))))
+identifier = Py.SimpleStatement (L1 (R1 (R1 (L1 (Py.ExpressionStatement () [L1 (L1 (Py.Expression (L1 (L1 (L1 (Py.PrimaryExpression (L1 (R1 (R1 (R1 (R1 (Py.Identifier () "hello"))))))))))))])))))
```
robrix: NB: The nesting structure is kinda hard to anticipate due to the balanced binary tree approach, but

1. we can abstract over it to project out parts of the tree, and
2. it’s really straightforward to do so—just a minor variation on the `Member` class `fused-effects` uses, really.

I’ve got a sketch of this somewhere on a branch that we can add to `semantic-ast` (or somewhere, anyway).

Plus, we measured a small but significant improvement to compile times by doing this; `ghc` has work to do proportional to the depth of the tree, more so than its arity, apparently. (This was a really cool hypothesis to validate!)

tree-sitter-python/test/Main.hs
```diff
@@ -31,7 +32,7 @@ prop_simpleExamples = property $ do
   "pass" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 pass] }
   "1" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 one] }
   "+1" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 plusOne] }
-  "expensive" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 function] }
+  "hello" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 identifier] }
```
robrix: Thank you for renaming this!

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -57,10 +56,10 @@ syntaxDatatype :: Ptr TS.Language -> Datatype -> Q [Dec]
...
   typeParameterName <- newName "a"
   case datatype of
-    SumType (DatatypeName datatypeName) _ subtypes -> do
-      cons <- traverse (constructorForSumChoice datatypeName typeParameterName) subtypes
-      result <- symbolMatchingInstanceForSums name subtypes
-      pure $ generatedDatatype name cons typeParameterName:result
+    SumType (DatatypeName _) _ subtypes -> do
+      types' <- fieldTypesToNestedSum subtypes
+      con <- normalC name [TH.bangType strictness (pure types' `appT` varT typeParameterName)]
+      pure [NewtypeD [] name [PlainTV typeParameterName] Nothing con [deriveGN, deriveStockClause, deriveAnyClassClause]]
```
robrix: I really love that we’re able to derive all these instances now ✨

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -92,21 +93,6 @@ symbolMatchingInstance language name str = do
...
-  where perMkType (MkType (DatatypeName n) named) = [e|TS.symbolMatch (Proxy :: Proxy $(conT (toName named n))) |]
-        mkOr lhs rhs = [e| (||) <$> $(lhs) <*> $(rhs) |]
-        extractn (MkType (DatatypeName n) Named) = toCamelCase n
-        extractn (MkType (DatatypeName n) Anonymous) = "Anonymous" <> toCamelCase n
-
-
--- | Append string with constructor name (ex., @IfStatementStatement IfStatement@)
-constructorForSumChoice :: String -> Name -> TreeSitter.Deserialize.Type -> Q Con
-constructorForSumChoice str typeParameterName (MkType (DatatypeName n) named) = normalC (toName named (n ++ str)) [child]
-  where child = TH.bangType strictness (appT (conT (toName named n)) (varT typeParameterName))
```
robrix: Yaaaas I love the reduction in surface area here so much ✨

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -170,7 +170,7 @@ class UnmarshalField t where
 instance UnmarshalField Maybe where
   unmarshalField []  = pure Nothing
   unmarshalField [x] = Just <$> unmarshalNode x
-  unmarshalField _   = fail "expected a node of type ((f :+: g) a) but got multiple"
+  unmarshalField _   = fail "expected a node of type (Maybe a) but got multiple"
```
robrix: 😁 thank you for fixing my typo!

tree-sitter-java/TreeSitter/Java/AST.hs
aymannadeem: it's easier to edit (add/remove), vertical lists are more readable to me than infinite left scroll.

robrix: Rad, thanks! (Maybe we should adopt this in `semantic` &c. too.)

PR #230, opened 2019-10-03
ghc 8.8.1

robrix:
This PR:

- [x] Enables a bunch of warnings everywhere.
- [x] Fixes a bunch of warnings.
- [x] Enables testing against ghc 8.8.1 in CI.

This doesn’t touch any version bounds (all of that happened in `semantic-source` & `semilattices`, both of which were minor versions), so this doesn’t require any new releases, altho it does resolve some harmless warnings which occur newly in ghc 8.8.1 by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ matrix:
   include:
     - compiler: "ghc-8.6.5"
       addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.5], sources: [hvr-ghc]}}
-    # - compiler: "ghc-8.8.1"
-    #   addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1], sources: [hvr-ghc]}}
+    - compiler: "ghc-8.8.1"
+      addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1], sources: [hvr-ghc]}}
 
 before_install:
 - mkdir -p $HOME/.local/bin
--- a/tree-sitter-go/tree-sitter-go.cabal
+++ b/tree-sitter-go/tree-sitter-go.cabal
@@ -11,20 +11,38 @@ copyright:           2017 GitHub
 category:            Tree-sitter, Go
 build-type:          Simple
 
+common common
+  default-language: Haskell2010
+  ghc-options:
+    -Weverything
+    -Wno-all-missed-specialisations
+    -Wno-implicit-prelude
+    -Wno-missed-specialisations
+    -Wno-missing-import-lists
+    -Wno-missing-local-signatures
+    -Wno-monomorphism-restriction
+    -Wno-name-shadowing
+    -Wno-safe
+    -Wno-unsafe
+  if (impl(ghc >= 8.6))
+    ghc-options: -Wno-star-is-type
+  if (impl(ghc >= 8.8))
+    ghc-options: -Wno-missing-deriving-strategies
+
 library
+  import: common
   exposed-modules:     TreeSitter.Go
   build-depends:       base              >= 4.12 && < 5
                      , tree-sitter       >= 0.3 && < 1
                      , template-haskell  >= 2.12 && < 2.16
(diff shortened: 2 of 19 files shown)


.travis.yml
```diff
@@ -22,8 +22,8 @@ matrix:
   include:
     - compiler: "ghc-8.6.5"
       addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.5], sources: [hvr-ghc]}}
-    # - compiler: "ghc-8.8.1"
-    #   addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1], sources: [hvr-ghc]}}
+    - compiler: "ghc-8.8.1"
+      addons: {apt: {packages: [cabal-install-3.0,ghc-8.8.1], sources: [hvr-ghc]}}
```
robrix: We do all of the checks against 8.8.1 now.

tree-sitter-go/tree-sitter-go.cabal
```diff
@@ -11,20 +11,38 @@ copyright:           2017 GitHub
...
+    -Wno-missing-import-lists
+    -Wno-missing-local-signatures
+    -Wno-monomorphism-restriction
+    -Wno-name-shadowing
+    -Wno-safe
+    -Wno-unsafe
+  if (impl(ghc >= 8.6))
+    ghc-options: -Wno-star-is-type
+  if (impl(ghc >= 8.8))
+    ghc-options: -Wno-missing-deriving-strategies
```
robrix: We enable these warnings in every package and `import: common` in every component.

tree-sitter-python/TreeSitter/Python/AST.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE DataKinds, DeriveAnyClass, DeriveGeneric, DeriveTraversable, DuplicateRecordFields, TemplateHaskell, TypeApplications #-}
-module TreeSitter.Python.AST where
+module TreeSitter.Python.AST
+( module TreeSitter.Python.AST
+) where
```
robrix: Explicit export lists are a good idea, but since we’re generating all of these with TH it’s easiest just to export everything declared in the module. (This is equivalent to omitting the explicit export list altogether.)

tree-sitter-python/test/Main.hs
```diff
@@ -1,23 +1,16 @@
...
-import           Control.Monad
 import           Control.Monad.IO.Class
 import           Data.Bool (bool)
 import           Data.ByteString (ByteString)
-import           Data.Char
-import           Data.Foldable
+import           GHC.Generics
 import           Hedgehog
-import qualified Hedgehog.Gen as Gen
-import qualified Hedgehog.Range as Range
```
robrix: Bunch of unused imports.

tree-sitter/src/TreeSitter/Deserialize.hs
```diff
@@ -3,6 +3,8 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE OverloadedStrings #-}
+-- Turn off partial field warnings for Datatype.
+{-# OPTIONS_GHC -Wno-partial-fields #-}
```
robrix: We use partial fields in `Datatype` and I didn’t want to change it.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -319,7 +319,7 @@ instance UnmarshalAnn k => GUnmarshal (K1 c k) where
 instance GUnmarshal Par1 where
   gunmarshalNode node = Par1 <$> unmarshalAnn node
 
-instance (Unmarshal t, SymbolMatching t) => GUnmarshal (Rec1 t) where
```
robrix: Redundant constraint 🔥

PR #232, opened 2019-10-04
Remove tree-sitter-haskell

rewinfrey:
Relates to https://github.com/tree-sitter/tree-sitter-haskell/issues/27, and temporarily removes `tree-sitter-haskell` to avoid compiling and building these modules until the Haskell grammar is in a good state again.


robrix:
I think we shouldn’t do this, because the package still exists on hackage either way.

rewinfrey:
@robrix ah, thanks for pointing that out. I'll close this.

PR #239, opened 2019-11-08
Python AST fixes for `import_from_statement`

tclem:
I noticed that python imports like `from foo import a, b, c` are not unmarshalling properly with an error `expected a node but didn't get one`. In debugging this, I've added a unit test, made the error messages a bit more informative and unique, and made a naming fix.

I'm now running into an issue in tree-sitter-python. `master` has the proper node-types.json for `import_from_statement`, but is missing a top level definition of `keyword_identifier`. If you use the following patch to manually fix `import_from_statement`, these tests pass.

``` diff
diff --git a/src/node-types.json b/src/node-types.json
index 2587a32..5825ca6 100644
--- a/src/node-types.json
+++ b/src/node-types.json
@@ -1370,8 +1370,8 @@
       }
     },
     "children": {
-      "multiple": true,
-      "required": true,
+      "multiple": false,
+      "required": false,
       "types": [
         {
           "type": "wildcard_import",
```

### DIFF
--- a/docs/codegen.md
+++ b/docs/codegen.md
@@ -28,10 +28,14 @@ The remaining document provides more details on generating ASTs, inspecting data
 ___
 
 ### Table of Contents
-- [Generating ASTs](#Generating-ASTs)
-- [Inspecting auto-generated datatypes](#Inspecting-auto-generated-datatypes)
-- [Tests](#tests)
-- [Additional notes](#Additional-notes)
+- [CodeGen Documentation](#codegen-documentation)
+    - [Prerequisites](#prerequisites)
+    - [CodeGen Pipeline](#codegen-pipeline)
+    - [Table of Contents](#table-of-contents)
+    - [Generating ASTs](#generating-asts)
+    - [Inspecting auto-generated datatypes](#inspecting-auto-generated-datatypes)
+    - [Tests](#tests)
+    - [Additional notes](#additional-notes)
 ___
 
 ### Generating ASTs
@@ -44,21 +48,21 @@ To parse source code and produce ASTs locally:
 cabal new-repl lib:tree-sitter-python
 ```
 
-2. Set language extensions, `OverloadedStrings` and `TypeApplications`, and import relevant modules, `TreeSitter.Unmarshal`, `TreeSitter.Range` and `TreeSitter.Span`:
+2. Set language extensions, `OverloadedStrings` and `TypeApplications`, and import relevant modules, `TreeSitter.Unmarshal`, `Source.Range` and `Source.Span`:
 
 ```
 :seti -XOverloadedStrings
 :seti -XTypeApplications
 
-import TreeSitter.Span
-import TreeSitter.Range
+import Source.Span
+import Source.Range
 import TreeSitter.Unmarshal
 ```
 
 3. You can now call `parseByteString`, passing in the desired language you wish to parse (in this case Python exemplified by `tree_sitter_python`), and the source code (in this case an integer `1`). Since the function is constrained by `(Unmarshal t, UnmarshalAnn a)`, you can use type applications to provide a top-level node `t`, an entry point into the tree, in addition to a polymorphic annotation `a` used to represent range and span:
 
 ```
-parseByteString @(TreeSitter.Python.AST.Module (TreeSitter.Range.Range, Span)) tree_sitter_python "1"
+parseByteString @TreeSitter.Python.AST.Module @(Source.Span.Span, Source.Range.Range) tree_sitter_python "1"
 ```
 
 This generates the following AST:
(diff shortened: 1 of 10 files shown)


docs/codegen.md
```diff
@@ -28,10 +28,14 @@ The remaining document provides more details on generating ASTs, inspecting data
...
-- [Tests](#tests)
-- [Additional notes](#Additional-notes)
+- [CodeGen Documentation](#codegen-documentation)
+    - [Prerequisites](#prerequisites)
+    - [CodeGen Pipeline](#codegen-pipeline)
+    - [Table of Contents](#table-of-contents)
+    - [Generating ASTs](#generating-asts)
+    - [Inspecting auto-generated datatypes](#inspecting-auto-generated-datatypes)
+    - [Tests](#tests)
+    - [Additional notes](#additional-notes)
```
aymannadeem: Thanks for updating docs!

docs/codegen.md
```diff
@@ -44,21 +48,21 @@ To parse source code and produce ASTs locally:
 cabal new-repl lib:tree-sitter-python
 ```
 
-2. Set language extensions, `OverloadedStrings` and `TypeApplications`, and import relevant modules, `TreeSitter.Unmarshal`, `TreeSitter.Range` and `TreeSitter.Span`:
+2. Set language extensions, `OverloadedStrings` and `TypeApplications`, and import relevant modules, `TreeSitter.Unmarshal`, `Source.Range` and `Source.Span`:
```
aymannadeem: Thanks for the fix, this was a remnant from before we decide to shuffle things around and keep them in `semantic-source`

tree-sitter-python/test/Main.hs
```diff
@@ -23,7 +24,14 @@ pass = Py.SimpleStatement (R1 (R1 (L1 (L1 (Py.PassStatement () "pass")))))
...
 identifier = Py.SimpleStatement (L1 (R1 (R1 (L1 (Py.ExpressionStatement () [L1 (L1 (Py.Expression (L1 (L1 (L1 (Py.PrimaryExpression (L1 (R1 (R1 (R1 (R1 (Py.Identifier () "hello"))))))))))))])))))
-
+fromImport = Py.SimpleStatement (R1 (L1 (L1 (R1 importStatement))))
+importStatement = Py.ImportFromStatement
+                    ()
+                    [ R1 (Py.DottedName () (Py.Identifier () "a" :| []))
+                    , R1 (Py.DottedName () (Py.Identifier () "b" :| []))
+                    ]
+                    (L1 (Py.DottedName () (Py.Identifier () "foo" :| [])))
+                    Nothing
```
aymannadeem: 💯 test coverage is improving slowly but surely

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -168,14 +160,6 @@ moduleForName n = Module . PkgName <$> namePackage n <*> (ModName <$> nameModule
...
-
--- Helper function to remove underscores from output of data type names
-removeUnderscore :: String -> String
-removeUnderscore = foldr appender ""
-  where appender :: Char -> String -> String
-        appender '_' cs = initUpper cs
-        appender c cs = c : cs
+-- | Replace operators and convert snake_case String to PascalCase
+toEscapedPascalCase :: String -> String
+toEscapedPascalCase = capitalize . escapeOperatorPunctuation . camelCase
```
aymannadeem: 👍

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -180,7 +181,7 @@ instance UnmarshalField NonEmpty where
     head' <- unmarshalNode x
     tail' <- unmarshalField xs
     pure $ head' :| tail'
-  unmarshalField [] = fail "expected a node but didn't get one"
+  unmarshalField [] = fail "expected a node of type (NonEmpty a) but got an empty list"
```
aymannadeem: 👍

tclem:
OK, @maxbrunsfeld made the right fixes in `tree-sitter-python` so this should be good to go now (once the build finishes).

docs/codegen.md
```diff
@@ -44,21 +48,21 @@ To parse source code and produce ASTs locally:
...
 
 ```
 :seti -XOverloadedStrings
 :seti -XTypeApplications
 
-import TreeSitter.Span
-import TreeSitter.Range
+import Source.Span
+import Source.Range
 import TreeSitter.Unmarshal
```
tclem: @aymannadeem what do you think about committing a `.ghci` file that does this?

aymannadeem: I have a `.ghci` file locally that takes care of this, happy to commit it since this is all pretty unusable without these key imports.

PR #240, opened 2019-11-14
Update tree sitter readme

aymannadeem:
Given that the library is now also concerned with CodeGen, the README felt a bit misleading. I re-route folks to the docs if they're interested in non-FFI stuff.


tree-sitter/README.md
```diff
@@ -1,6 +1,10 @@
 # haskell-tree-sitter
 
-**Please note that this library is currently in flux. The maintainers are revising the core API to allow for generation of Haskell types from a tree-sitter grammar, and the automatic bridging of tree-sitter syntax nodes to these types.**
```
aymannadeem: I think we can say it's stable enough to not warrant this message anymore? Happy leaving it in there too, given the precise AST explorations in Semantic may bring some new insights about what needs to change. Nothing should be major, though.

tree-sitter/README.md
```diff
@@ -1,6 +1,10 @@
...
+There are two parts to this library:
+1. [Haskell bindings for tree-sitter](#haskell-bindings-for-tree-sitter)
+2. [CodeGen API](#codegen-api)
```
aymannadeem: This library does a lot more now!

XVilka:
Should be this merged then? Since was approved and makes a perfect sense.

PR #241, opened 2019-11-14
Reuse tree-sitter's corpus examples to assert new AST unmarshalling

tclem:
Takes the existing tree-sitter-python corpus examples, which are snippets of python code used to assert the tree-sitter grammar, and sends those code snippets through the code generated by `node-type.json` to assert unmarshalling and provide some confidence of end-to-end functionality.

NOTE: There are a number of failing examples here, so I don't expect the build to be green. We will have to dig into what's going on and make some fixes before this is ready to merge. Here's a paste of the current output for context:

```
Running 1 test suites...
Test suite test: RUNNING...
Tests
  tree-sitter corpus tests
    ./vendor/tree-sitter-python/corpus/literals.txt
      Integers:                                  OK
      Floats:                                    OK
      Scientific Notation Floats:                OK
      Strings:                                   OK
      Raw strings:                               OK
      Raw strings with escaped quotes:           OK
      Format strings:                            OK
      Format strings with format specifiers:     OK
      Unicode escape sequences:                  OK
      Other primitives:                          OK
      Concatenated strings:                      OK
      Multi-line strings:                        OK
      Lists:                                     OK
      List comprehensions:                       FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        [a + b for (a, b) in items]
        [a for b in c for a in b]
        [(x,y) for x in [1,2,3] for y in [1,2,3] if True]
        [a for a in lambda: True, lambda: False if a()]
        ```
      Dictionaries:                              OK
      Dictionary comprehensions:                 OK
      Sets:                                      OK
      Set comprehensions:                        OK
      Tuples:                                    OK
      Generator expression:                      OK
    ./vendor/tree-sitter-python/corpus/statements.txt
      Import statements:                         OK
      Import-from statements:                    OK
      Future import statements:                  OK
      Print statements:                          OK
      Print statements with redirection:         OK
      Assert statements:                         OK
      Expression statements:                     OK
      Delete statements:                         OK
      Control-flow statements:                   OK
      Return statements:                         OK
      If statements:                             OK
      If else statements:                        OK
      Nested if statements:                      OK
      While statements:                          OK
      For statements:                            OK
      Try statements:                            OK
      With statements:                           OK
      Async Function definitions:                FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        async def a():
          b

        async def c(d):
          e

        async def g(g, h,):
          i

        async def c(a: str):
          a

        async def c(a: b.c):
          a

        async def d(a: Sequence[T]) -> T:
          a

        async def i(a, b=c, *c, **d):
          a

        async def d(a: str) -> None:
          return None

        async def d(a:str="default", b=c) -> None:
          return None
        ```
      Function definitions:                      FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        def e((a,b)):
          return (a,b)

        def e(*list: str):
          pass

        def e(**list: str):
          pass

        def f():
          nonlocal a

        def g(h, i, *, j, k=100, **kwarg):
          return h,i,j,k,kwarg

        def h(*a):
          i((*a))
          j(((*a)))
        ```
      Class definitions:                         OK
      Class definitions with superclasses:       OK
      Decorated definitions:                     OK
      Raise statements:                          OK
      Comments:                                  OK
      Comments at different indentation levels:  OK
      Comments after dedents:                    OK
      Newline tokens followed by comments:       OK
      Global statements:                         OK
      Exec statements:                           OK
      Extra newlines:                            OK
    ./vendor/tree-sitter-python/corpus/expressions.txt
      Identifiers with Greek letters:            OK
      Subscript expressions:                     OK
      Subscript slice expressions:               OK
      Attribute references:                      OK
      Await expressions:                         FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        await i(j, 5)
        return await i(j, 5)
        ```
      Call expressions:                          OK
      Print used as an identifier:               OK
      Print used as a parameter:                 OK
      Exec used as an identifier:                OK
      Calls with splats:                         OK
      Math operators:                            OK
      Binary Addition / Subtraction With Floats: OK
      Bitwise operators:                         OK
      Boolean operators:                         OK
      Comparison operators:                      OK
      Assignments:                               OK
      Assignments with type annotations:         OK
      Augmented assignments:                     OK
      Named expressions:                         FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        a := x
        (y := f(x))
        foo(x=(y := f(x)))
        y0 = (y1 := f(x))
        def foo(answer=(p := 42)):
          return answer;
        def foo(answer: (p := 42) = 5):
          return answer;
        foo(x := 3, cat='vector')
        (z := (y := (x := 0)))
        ```
      Yield expressions:                         FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        def example():
          yield
          yield 1
          x = yield 2
          yield from a
          yield from (yield from (x for x in range(1, 10)))
        ```
      lambdas:                                   FAIL
        test/Test.hs:42:
        failed to parse:
        ``` python
        lambda b, c: d("e" % f)
        lambda: True
        lambda a, b = c, *d, **e: a
        lambda (a,b): (a, b)
        ```
      Conditional if expressions:                OK
      Async context managers and iterators:      OK
  (checked by Hedgehog)
    prop_simpleExamples:                         OK (0.04s)
        ✓ prop_simpleExamples passed 100 tests.

7 out of 74 tests failed (0.04s)
Test suite test: FAIL
Test suite logged to:
/Users/tclem/github/haskell-tree-sitter/dist-newstyle/build/x86_64-osx/ghc-8.6.5/tree-sitter-python-0.6.0.0/t/test/test/tree-sitter-python-0.6.0.0-test.log
0 of 1 test suites (0 of 1 test cases) passed.
cabal: Tests failed for test:test from tree-sitter-python-0.6.0.0.
```


tree-sitter-python/test/Test.hs
```diff
@@ -0,0 +1,76 @@
...
+corpusParser = do
+  xs <- many' exampleParser
+  void endOfInput
+  pure xs
+
+exampleParser :: Parser CorpusExample
+exampleParser = do
+  name <- exampleNameParser
+  code <- manyTill anyChar (string "\n---\n")
+  _out <- manyTill anyChar (choice [endOfInput, char '=' $> ()])
```
tclem: We don't capture or assert sexpression output, but we could.

tree-sitter-python/test/Test.hs
```diff
@@ -0,0 +1,76 @@
...
+
+data CorpusExample = CorpusExample { name :: String, code :: ByteString }
+  deriving (Eq, Show)
+
+parseCorpusFile :: Path.RelFile -> IO (Either String [CorpusExample])
+parseCorpusFile path = do
+  c <- Data.ByteString.readFile (Path.toString path)
+  pure $ parseOnly corpusParser c
+
+corpusParser :: Parser [CorpusExample]
```
tclem: This is a pretty quick pass at parsing, feedback welcome as I'm sure there are more concise ways to express these parser combinators.

tree-sitter-python/test/Test.hs
```diff
@@ -0,0 +1,76 @@
...
+testCorpus path = do
+  xs <- parseCorpusFile path
+  case xs of
+    Left e -> print ("failed to parse corpus: " <> show path) *> print e *> exitFailure
+    Right xs -> testGroup (Path.toString path) <$> traverse toPropTest xs
+  where
+    toPropTest (CorpusExample name code) = do
+      res <- parseByteString @Py.Module @() tree_sitter_python code
+      let failureMsg = "failed to parse:\n``` python\n" <> code <> "```"
+      pure $ testCase name (assertBool (unpack failureMsg) (isRight res))
```
tclem: We just assert that the code example parses and unmarshalls OK (`isRight`).

aymannadeem:
Still looking into this. Testing in the REPL earlier, I confirmed we're definitely able to generate Template Haskell datatypes based on the `node-types.json` file for `lambda`, `await` and `yield`, for ex:

```Haskell
data TreeSitter.Python.AST.Lambda a
  = TreeSitter.Python.AST.Lambda {TreeSitter.Python.AST.ann :: a,
                                  TreeSitter.Python.AST.body :: (GHC.Generics.:+:)
                                                                  TreeSitter.Python.AST.Expression
                                                                  TreeSitter.Python.AST.Lambda
                                                                  a,
                                  TreeSitter.Python.AST.parameters :: Maybe
                                                                        (TreeSitter.Python.AST.LambdaParameters
                                                                           a)}

```

However, when we try to parse the bytestring (and go through the unmarshal code path) to bridge the type to the source, it fails:

``` parseByteString @TreeSitter.Python.AST.Module @(Source.Span.Span, Source.Range.Range) tree_sitter_python "lambda: True"
Left "Expected Attribute but got Lambda. Expected BinaryOperator but got Lambda. Expected Call but got Lambda. Expected ConcatenatedString but got Lambda. Expected Dictionary but got Lambda. Expected DictionaryComprehension but got Lambda. Expected Ellipsis but got Lambda. Expected False but got Lambda. Expected Float but got Lambda. Expected GeneratorExpression but got Lambda. Expected Identifier but got Lambda. Expected Integer but got Lambda. Expected List but got Lambda. Expected ListComprehension but got Lambda. Expected None but got Lambda. Expected ParenthesizedExpression but got Lambda. Expected Set but got Lambda. Expected SetComprehension but got Lambda. Expected String but got Lambda. Expected Subscript but got Lambda. Expected True but got Lambda. Expected Tuple but got Lambda. Expected UnaryOperator but got Lambda. Expected Await but got Lambda. Expected BooleanOperator but got Lambda. Expected ComparisonOperator but got Lambda. Expected ConditionalExpression but got Lambda. Expected Lambda but got Lambda. Expected NamedExpression but got Lambda. Expected NotOperator but got Lambda. Expected Assignment but got Lambda. Expected AugmentedAssignment but got Lambda. Expected Yield but got Lambda"
*TreeSitter.Python TreeSitter.Unmarshal Source.Span Source.Range>
```

The `symbolMatchingInstance` is being invoked to create an instance, in this case for products and I think we're getting the right `TSSymbol` from the node, but expect keywords that have more meaning than a string identifier. 

Not sure if this is right but we might wanna do something similar to [this](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/tree-sitter/src/TreeSitter/Unmarshal/Examples.hs), where we hardcode certain datatypes and have SymbolMatching instances for each. I got a bit distracted with other stuff today but can look into that tomorrow.

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -87,7 +87,7 @@ symbolMatchingInstance language name str = do
   tsSymbol <- runIO $ withCString str (TS.ts_language_symbol_for_name language)
   tsSymbolType <- toEnum <$> runIO (TS.ts_language_symbol_type language tsSymbol)
   [d|instance TS.SymbolMatching $(conT name) where
-      showFailure _ node = "Expected " <> $(litE (stringL (show name))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
+      showFailure _ node = "Expected " <> $(litE (stringL (TS.symbolToName tsSymbolType str))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
```
aymannadeem: There are _two_ of all the language constructs we're trying to parse (`yeild` is `AnonymousYield` the anonymous leaf type and also `Yield` the product type). We're for some reason expecting the leaf when we actually should expect the product, causing us to fail.

aymannadeem:
When we encounter the situation in which an anonymous leaf shares a common name with a product, we (correctly) parse the product node, but the call to `symbolToName` `(TS.symbolToName tsSymbolType str)` results in attempting to parse the leaf.

```
*TreeSitter.Python TreeSitter.Unmarshal Source.Span Source.Range> parseByteString @TreeSitter.Python.AST.Module @(Source.Span.Span, Source.Range.Range) tree_sitter_python "yield 1"
Left "Expected Attribute but got Yield. Expected BinaryOperator but got Yield. Expected Call but got Yield. Expected ConcatenatedString but got Yield. Expected Dictionary but got Yield. Expected DictionaryComprehension but got Yield. Expected Ellipsis but got Yield. Expected False but got Yield. Expected Float but got Yield. Expected GeneratorExpression but got Yield. Expected Identifier but got Yield. Expected Integer but got Yield. Expected List but got Yield. Expected ListComprehension but got Yield. Expected None but got Yield. Expected ParenthesizedExpression but got Yield. Expected Set but got Yield. Expected SetComprehension but got Yield. Expected String but got Yield. Expected Subscript but got Yield. Expected True but got Yield. Expected Tuple but got Yield. Expected UnaryOperator but got Yield. Expected AnonAwait but got Yield. Expected BooleanOperator but got Yield. Expected ComparisonOperator but got Yield. Expected ConditionalExpression but got Yield. Expected AnonLambda but got Yield. Expected NamedExpression but got Yield. Expected NotOperator but got Yield. Expected Assignment but got Yield. Expected AugmentedAssignment but got Yield. Expected AnonYield but got Yield"
```

Expect `blah` but got `Yield` means that `blah` are all of the possible things that can happen before, in this case, `1`.

tree-sitter-python/test/Manual/Examples.hs
```diff
@@ -1,18 +1,27 @@
...
 import qualified TreeSitter.Python.AST as Py
 import           TreeSitter.Token
 import           TreeSitter.Unmarshal
 
+tests :: TestTree
+tests = testGroup "(checked by Hedgehog)" props
+  where
+    props = let (Group _ xs) = $$(discover)
+            in fmap toP xs
+    toP ((PropertyName name), prop) = Hedgehog.testProperty name prop
```
tclem: I did this to keep around the existing property tests. These might be redundant now. What do you think @aymannadeem and @maxbrunsfeld - should we remove them entirely?

tree-sitter-python/.ghci
```diff
@@ -5,3 +5,4 @@
 import TreeSitter.Unmarshal
 import Source.Span
 import Source.Range
+:set prompt "> "
```
tclem: @aymannadeem is this OK? I like having a short ghci prompt without all the module names.

tree-sitter-python/test/Test.hs
tclem: In the next PR I'll extract this into a common testing library so that the other languages can also use this testing strategy.

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -44,29 +41,42 @@ astDeclarationsForLanguage language filePath = do
...
   let invocationRelativePath = takeDirectory (pwd </> currentFilename) </> filePath
   input <- runIO (eitherDecodeFileStrict' invocationRelativePath)
-  either fail (fmap (concat @[]) . traverse (syntaxDatatype language)) input
-
+  allSymbols <- runIO (getAllSymbols language)
+  either fail (fmap (concat @[]) . traverse (syntaxDatatype language allSymbols)) input
+
+-- Build a list of all symbols
+getAllSymbols :: Ptr TS.Language -> IO [(String, Named)]
+getAllSymbols language = do
```
tclem: This replaces the `TS. ts_language_symbol_for_name` API entirely as we were having issues with there being multiple symbols for a name. Instead, we build a (compile time) list of symbols up front so that we can lookup `TSSymbol`s later.

tree-sitter/src/TreeSitter/Symbol.hs
```diff
@@ -1,95 +1,133 @@
...
-, symbolToName
-, escapeOperatorPunctuation
-) where
+  ( TSSymbol
+  , fromTSSymbol
+  , SymbolType(..)
+  , Symbol(..)
+  , symbolToName
+  , toHaskellCamelCaseIdentifier
+  , toHaskellPascalCaseIdentifier
```
tclem: I moved all the functions that are handling the mapping of names in here, along with a couple of string helpers that are now only used directly in the tests.

tree-sitter/src/TreeSitter/Unmarshal.hs
```diff
@@ -247,7 +246,7 @@ peekFieldName = do
   if fieldName == nullPtr then
     pure Nothing
   else
-    Just . FieldName . camelCase <$> liftIO (peekCString fieldName)
+    Just . FieldName . toHaskellCamelCaseIdentifier <$> liftIO (peekCString fieldName)
```
tclem: This was an important fix for reserved names in haskell like `type` which get a prime appended to them.

tree-sitter-python/.ghci
aymannadeem: yes, I wasn't sure if it was ok to be this prescriptive with other people's prompts. 😄

tree-sitter-python/test/Test.hs
aymannadeem: That would be great!

tree-sitter-python/test/Manual/Examples.hs
maxbrunsfeld: I'm not sure what properties they test. Is there overlap with what we're verifying here?

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -82,14 +92,13 @@ syntaxDatatype language datatype = skipDefined $ do
 
 
 -- | Create TH-generated SymbolMatching instances for sums, products, leaves
-symbolMatchingInstance :: Ptr TS.Language -> Name -> String -> Q [Dec]
-symbolMatchingInstance language name str = do
-  tsSymbol <- runIO $ withCString str (TS.ts_language_symbol_for_name language)
-  tsSymbolType <- toEnum <$> runIO (TS.ts_language_symbol_type language tsSymbol)
+symbolMatchingInstance :: [(String, Named)] -> Name -> Named -> String -> Q [Dec]
+symbolMatchingInstance allSymbols name named str = do
+  let tsSymbols = elemIndices (str, named) allSymbols
```
maxbrunsfeld: Since we're going to be calling `symbolMatch` a lot, I wonder if there's some other data structure that we should be using for `tsSymbols` instead of a list, like a `Set` or something, which might have a faster `elem`.

Usually these lists will only have 1 element, and only occasionally will they have two or three, so maybe a list is fine. Just wanted to raise this in case anyone more familiar w/ haskell data structures has a preference here.

tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -82,14 +92,13 @@ syntaxDatatype language datatype = skipDefined $ do
...
+  let names = intercalate ", " $ fmap (fst . (!!) allSymbols) tsSymbols
   [d|instance TS.SymbolMatching $(conT name) where
-      showFailure _ node = "Expected " <> $(litE (stringL (show name))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
-      symbolMatch _ node = TS.fromTSSymbol (nodeSymbol node) == $(conE (mkName $ "Grammar." <> TS.symbolToName tsSymbolType str))|]
-
+      showFailure _ node = "expected " <> $(litE (stringL (show names))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
```
maxbrunsfeld: Currently, this will have a different naming convention for the `expected` than for the `got`, right? Like `Expected yield_expression but got AwaitExpression`. Or is that fixed? I think ideally, we'd drop the auto-generated enum entirely, and say `Expected yield_expression but got await_expression`.

Either way, I don't think it needs to block this PR necessarily.

tree-sitter-python/test/Manual/Examples.hs
aymannadeem: I can't quite tell from this snippet what the delta between the other tests vs. these properties are, but I've found property tests super helpful in shedding light on blind spots historically 🤷‍♀

tree-sitter/src/TreeSitter/GenerateSyntax.hs
aymannadeem: tbqh I'm not sure either, maybe @patrickt can chime in?

tree-sitter/src/TreeSitter/GenerateSyntax.hs
tclem: Yeah we could make this better, but let's do it in another PR. I got a bit confused with the compile time and template haskell vs. runtime stuff here to get the error messages right.

tree-sitter-python/test/Manual/Examples.hs
tclem: If you expand the context 👇 below you can see. They aren't actually property tests, more like unit tests just using the hedgehog `Property`. It's stuff like:

``` haskell
"hello" `shouldParseInto` Py.Module { Py.ann = (), Py.extraChildren = [R1 identifier] }
```

... which is why I think it all might be covered by the corpus tests, but let's leave it for now and clean up in another pass.

tree-sitter-python/.ghci
robrix: @tclem: IMO this belongs in your user .ghci file, not in any particular project’s.

robrix: E.g. here’s my ridiculous prompt: https://github.com/robrix/dotfiles/blob/master/home/.ghc/ghci.conf#L23-L34

tclem: Ah, right on. Feel free to remove this the next time we're changing these files.

PR #243, opened 2019-11-20
Test precise ASTs against tree-sitter-java corpus

tclem:
Now that https://github.com/tree-sitter/haskell-tree-sitter/pull/241 landed we can use this same strategy to assert precise ASTs in other languages. This PR extracts a package for some common test functions and wires up testing of the tree-sitter-java corpus for precise ASTs.


tree-sitter-java/.ghci
```diff
@@ -0,0 +1 @@
+../.ghci
```
tclem: Just a symlink into each project

tclem:
I almost forgot one very important note: java pretty much passes on the first try here! ✨

PR #246, opened 2019-11-20
Version bumps for releasing tree-sitter and a couple of language packages

tclem:


tclem:
Released:

- https://hackage.haskell.org/package/tree-sitter-0.6.0.0
- https://hackage.haskell.org/package/tree-sitter-java-0.5.0.0
- https://hackage.haskell.org/package/tree-sitter-python-0.7.0.0
- https://hackage.haskell.org/package/tree-sitter-json-0.4.0.0

PR #248, opened 2019-11-21
Better symbol matching errors

tclem:
Improves the error messages in generated `SymbolMatching` instances.`showFailure` will display the symbol names as known to tree-sitter (no string manipulation).


tree-sitter/src/TreeSitter/GenerateSyntax.hs
```diff
@@ -95,11 +90,16 @@ syntaxDatatype language allSymbols datatype = skipDefined $ do
...
+  let names = intercalate ", " $ fmap (debugPrefix . (!!) allSymbols) tsSymbols
   [d|instance TS.SymbolMatching $(conT name) where
-      showFailure _ node = "expected " <> $(litE (stringL (show names))) <> " but got " <> show (TS.fromTSSymbol (nodeSymbol node) :: $(conT (mkName "Grammar.Grammar")))
+      showFailure _ node = "expected " <> $(litE (stringL (show names))) <> " but got " <> show (debugPrefix (allSymbols !! fromIntegral (nodeSymbol node)))
       symbolMatch _ node = elem (nodeSymbol node) tsSymbols|]
 
+-- | Prefix symbol names for debugging to disambiguate between Named and Anonymous nodes.
+debugPrefix :: (String, Named) -> String
+debugPrefix (name, Named)     = name
+debugPrefix (name, Anonymous) = "_" <> name
```
aymannadeem: good call

aymannadeem (approved):

maxbrunsfeld:
This is great. I think it means that for the most part, the precise AST code no longer depends on the older auto-generated `enum` of all of the symbols.

PR #249, opened 2019-11-25
Json marshal

aymannadeem:
Starting to do the work to generically serialize the generically unmarshaled ASTs into JSON format. This will (I think) entail marshaling those ASTs and auto-deriving `Aeson` and a few other typeclass instances in `GenerateSyntax` in order to be able to call `marshal` in `semantic-ast`.


aymannadeem:
@robrix assuming it's safe to close this out in lieu of https://github.com/github/semantic/pull/384?

robrix:
@aymannadeem yep 👍

PR #255, opened 2019-12-12
Enable precise ASTs for a few more languages

tclem:
This enables precise ASTs for:

- [x] Ruby
- [x] Go
- [x] TypeScript
- [x] TSX
- [ ] ~~PHP~~

EDIT: Will follow up with PHP support in another branch.


tree-sitter-ruby/.ghci
```diff
@@ -0,0 +1 @@
+../.ghci
```
robrix: This is a symlink, yeah? And not e.g. a regular file with these contents?

tree-sitter/src/TreeSitter/Symbol.hs
```diff
@@ -114,11 +114,12 @@ escapeOperatorPunctuation = concatMap $ \case
...
-camelCase = foldr appender mempty
+camelCase = go
   where
-    appender :: Char -> String -> String
-    appender '_' cs = capitalize cs
-    appender c cs   = c : cs
+    go ('_':'_':xs) = "Underscore" <> go xs
+    go ('_':xs)     = go (capitalize xs)
+    go (x:xs)       = x : go xs
+    go ""           = ""
```
robrix: Will this require any kind of version bump?

tclem: We'll have to release new versions of tree-sitter and tree-sitter-$lang to get all this stuff out. I don't think this is a break change as we hadn't yet tried to generate syntax for things like `__END__` in Ruby/Pearl and the existing generated syntax for Python/Java won't change (it's only double underscore that has to be different than just a single underscore).

tree-sitter-ruby/.ghci
tclem: Yup. symlink.

tclem:
Published to hackage.

PR #259, opened 2020-01-14
Release tree-sitter 0.8.0.0 and minor versions of select language packages

tclem:


tclem:
These have been released to hackage.

PR #260, opened 2020-01-22
Interruptible copies from cursors

robrix:
We think we’re seeing some copies that we would like to interrupt.

### DIFF
--- a/tree-sitter/src/TreeSitter/Cursor.hs
+++ b/tree-sitter/src/TreeSitter/Cursor.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE InterruptibleFFI #-}
 module TreeSitter.Cursor
 ( Cursor
 , withCursor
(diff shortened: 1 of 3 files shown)


tree-sitter/src/TreeSitter/Cursor.hs
```diff
@@ -51,4 +51,4 @@ foreign import ccall unsafe "ts_tree_cursor_goto_next_sibling" ts_tree_cursor_go
 foreign import ccall unsafe "ts_tree_cursor_goto_first_child" ts_tree_cursor_goto_first_child :: Ptr Cursor -> IO Bool
 foreign import ccall unsafe "ts_tree_cursor_goto_first_child_for_byte" ts_tree_cursor_goto_first_child_for_byte :: Ptr Cursor -> Word32 -> IO Int64
 
-foreign import ccall unsafe "src/bridge.c ts_tree_cursor_copy_child_nodes" ts_tree_cursor_copy_child_nodes :: Ptr Cursor -> Ptr Node -> IO Word32
+foreign import ccall interruptible "src/bridge.c ts_tree_cursor_copy_child_nodes" ts_tree_cursor_copy_child_nodes :: Ptr Cursor -> Ptr Node -> IO Word32
```
robrix: This matches what we do for copies from a node.

robrix (commented):
Ready for review.

tclem (approved):

robrix:
NB: #261 applies to this as well, tho not as badly as to `ts_node_copy_child_nodes`.

robrix:
https://hackage.haskell.org/package/tree-sitter-0.8.0.1

PR #262, opened 2020-01-23
Bump tree-sitter & tree-sitter-go versions

### LINKED ISSUE #261: Interrupting the copy_child_nodes functions can leak memory
Because we’re using `-XInterruptibleFFI` for `ts_node_copy_child_nodes` (and `ts_tree_cursor_copy_child_nodes`, as of #260), the runtime will interrupt the C function by killing the thread. However, these move the cursor, which allocates memory to manage the cursor’s stack; and `ts_node_copy_child_nodes` allocates a cursor as well.

Thus, any time we interrupt one of these calls, we’re leaking memory, because neither of these functions installs a signal handler to clean up their allocations.

robrix:
- [x] Fixes #261.

### DIFF
--- a/tree-sitter-go/tree-sitter-go.cabal
+++ b/tree-sitter-go/tree-sitter-go.cabal
@@ -1,6 +1,6 @@
 cabal-version:       2.4
 name:                tree-sitter-go
-version:             0.4.1.0
+version:             0.4.1.1
 synopsis:            Tree-sitter grammar/parser for Go
 description:         This package provides a parser for Go suitable for use with the tree-sitter package.
 license:             BSD-3-Clause
--- a/tree-sitter/ChangeLog.md
+++ b/tree-sitter/ChangeLog.md
@@ -1,6 +1,7 @@
 ### v0.8.0.2
 
 * Updates `tree-sitter` to fix an issue with improperly balanced subtrees leading to slowdowns when copying nodes.
+* Neither `ts_tree_cursor_copy_child_nodes` nor `ts_node_copy_child_nodes` is interruptible.
 
 
 ### v0.8.0.1
--- a/tree-sitter/ChangeLog.md
+++ b/tree-sitter/ChangeLog.md
@@ -1,3 +1,8 @@
+### v0.8.0.2
+
+* Updates `tree-sitter` to fix an issue with improperly balanced subtrees leading to slowdowns when copying nodes.
+
+
 ### v0.8.0.1
 
 * `ts_tree_cursor_copy_child_nodes` is interruptible.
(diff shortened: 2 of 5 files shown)


robrix:
https://hackage.haskell.org/package/tree-sitter-go-0.4.1.1 is live (meant to get review before that, my bad)

robrix:
Candidate of tree-sitter 0.8.0.2: https://hackage.haskell.org/package/tree-sitter-0.8.0.2/candidate

robrix:
https://hackage.haskell.org/package/tree-sitter-0.8.0.2 is live

maxbrunsfeld:
A little post-mortem:

In the past few days, we've been trying to upgrade GitHub's code-navigation service to use the newest version of Semantic. But we were seeing significant slowdowns when trying to deploy the new version. This was very surprising to us, because the new version of Semantic was showing much better performance in our benchmarks.

### Isolating the Bug

Yesterday, @tclem and @robrix were able to find an example of a file that was triggering the problem: [this generated Go file](https://github.com/grafana/grafana/blob/master/vendor/github.com/rainycape/unidecode/table.go) which contains a string literal containing thousands of escape sequences.

It turned out that this file was triggering a performance bug in Tree-sitter: parsing would complete successfully, and very quickly, but the internal structure of the resulting tree would be severely unbalanced:

unbalanced-tree

For strings with *very* large numbers of escape sequences (> 1000), this unbalanced structure will impose a major slowdown in any Tree-sitter API function that must retrieve a syntax node's *parent* node.

### What Changed

Interestingly, this is *not* a new Tree-sitter bug; it has always existed. What *is* new is that, when walking the Tree-sitter tree in Haskell to build up Semantic's langauge-specific abstract syntax tree, we had begun to use a Tree-sitter API function called [`ts_tree_cursor_get_field_name`](https://github.com/tree-sitter/tree-sitter/blob/9ffcb16392941c15fdd8bf45c8ea6ea3f1799668/lib/include/tree_sitter/api.h#L613-L619). The tree's lack of balance made that function take a very long time.

### The Fix

@robrix @patrickt and I found the problem in Tree-sitter: there was a mistake in the code that was opting *out* of the tree-balancing procedure because the children of the unbalanced nodes were single tokens (as opposed to internal nodes with some additional structure within them). I don't know why that condition was there. In essence, we removed that bogus condition, and allowed the tree-balancing to kick in:

balanced-tree

After deploying this to production, we're seeing much better performance. 🎊

robrix:
At the same time, we had expected that our use of `InterruptibleFFI` should be causing `tree-sitter` to interrupt the thread calling `ts_tree_cursor_get_field_name`. But while that does result in `pthread_kill` signalling the thread, this relies on the signal being observed—if you don’t check `errno` for `EINTR`, you’re going to continue to spin away, exactly as we observed. We’ve thus removed the use of `InterruptibleFFI`, as the fix to `tree-sitter` makes it redundant anyway.

PR #265, opened 2020-02-03
Clean up codegen

### LINKED ISSUE #423: AST datatypes should live in semantic
It’s agonizing trying to work on anything in `semantic` that involves touching `haskell-tree-sitter`, not least because of the hackage releases eventually involved for each of the `tree-sitter*` packages.

Further to this, there’s the `cabal` bug with internal libraries that keeps breaking our local builds every so often; the general painfulness of submodules; we’re really not ready for a hackage release yet; and the weird situation where `semantic-source` is depended upon by `tree-sitter` so it can’t be listed in `cabal.project` and that makes changing _it_ harder.

All of this is only going to get worse as we support more languages, too.

Given all of that, I’m of the opinion that the AST datatypes should live in one or more `semantic` packages. This would mean that:

- the `tree-sitter-*` packages would basically just build the parsers & vend the `tree_sitter_foo` symbol;
- the `semantic-*` language packages will contain the ASTs; and
- we would move codegen and unmarshalling into a semantic package, probably `semantic-ast`.

Some knock-on effects:

- the `tree-sitter-*` packages would really only change when we wanted to bump to the latest versions of the parsers; hackage releases aren’t a huge deal in this case
- changing the AST types, instances, etc. would happen entirely in semantic
- we could iterate more easily on e.g. performance of stuff like unmarshalling (which we were doing a lot of recently)
- there’d be _zero_ internal libraries to worry about, solving that pain once and for all
- `semantic-source` could be part of the `cabal.project` again instead of living in a weird little universe all by itself and not being very easy to work on
- I could add `Traversable1` to `semantic-ast` and derive instances of it for our ASTs without having to release `semantic-ast` on hackage (which it is much too early for)

aymannadeem:
Fixes https://github.com/github/semantic/issues/423. This PR is the other side of https://github.com/github/semantic/pull/448 and https://github.com/github/semantic/pull/468; all codegen files moved into `Semantic` will be deleted here.

TODO:
- [x] Delete `tree-sitter-LANGUAGE/LANGUAGE` containing AST + Grammar datatype gen code for Go, Ruby, Python, TypeScript, TSX, JSON, Java, 
- [x] Delete `Deserialize`
- [x] Delete `Token`
- [x] Delete `GenerateSyntax`
- [x] Delete `Unmarshal`
- [x] Update `tree-sitter.cabal` by removing exposed modules
- [x] Update `tree-sitter-LANGUAGE.cabal` files and remove internal libraries entirely
- [x] Migrate `test/Test.hs` files for all langs and `TreeSitter.Test.Helpers` to `semantic-codegen` via https://github.com/github/semantic/pull/468


patrickt (approved):
This looks great! Amazing work—once the package versions are bumped (which you can do here or elsewhere), we’ll do a full Hackage release, and we’ll never have to hit the dreaded cabal bug again!

joshvera (approved):
🚢

aymannadeem:
Did Hackage releases for `tree-sitter` + all language packages.

PR #266, opened 2020-02-11
Patch release to expose php node-types.json

aymannadeem:
Fixes https://github.com/github/semantic/pull/468 

Experiencing [a CI failure in Semantic](https://github.com/github/semantic/pull/468/checks?check_run_id=439636879) currently resulting from the `data-files` not exposing the `node-type.json` file required. 

This is a patch to fix that and release v0.3.0.1 for `tree-sitter-php`


aymannadeem:
Once this is merged, we should see some errors over in https://github.com/github/semantic/pull/468 disappear. The PHP `node-types.json` was not being picked up by the call to `getNodeTypes` in Semantic. This had to be exposed in `data-files`, followed by a hackage release dance.

joshvera (approved):

maxbrunsfeld:
I remember we were holding off for a while on bumping the PHP parser in semantic because of some big changes to the grammar that might require a lot of changes in assignment. Not sure if that is still anything to be concerned about.

tclem:
We just need to do the assignment fixes in semantic if we're going to update tree-sitter-php (which is a little bit of sunk cost effort since assignment is going away).

aymannadeem:
@tclem I was trying to fix Assignment errors I'm seeing in my other branch on Semantic. Agreed it's not ideal but I wasn't sure if it was okay to jump the gun and deprecate PHP Assignment just yet. I wanted to be more incremental while fixing those CI errors I'm seeing.

PR #271, opened 2020-03-04
emendation of Demo.hs link

p-alik:
This aims to improve [tree-sitter/README.md](../blob/master/tree-sitter/README.md)


Kleidukos:
fixed by #299

PR #272, opened 2020-03-05
Update tree-sitter-ql to use consistent path to test files

rewinfrey:
While adding the `tree-sitter-ql` package to `semantic`, I noticed that `tree-sitter-ql` had a top-level `corpus` directory rather than the standard `test/corpus` directory structure found in most tree-sitter projects.

This bumps the vendored `tree-sitter-ql` parser and corpus files so that `semantic` can use the standard test file path.


aymannadeem (approved):
👍🏽

PR #275, opened 2020-04-02
Update tree-sitter-ql

rewinfrey:
This pulls in changes to the `tree-sitter-ql` parser and introduces `name` field nodes for module expressions.

A new [hackage package](https://hackage.haskell.org/package/tree-sitter-ql-0.1.0.3) has been published.


patrickt (approved):
Boom.

PR #282, opened 2020-06-15
Add 8.10.1 support for haskell tree-sitter packages

joshvera:
Adds 8.10.1 support by pinning `text` and `fused-effects` and updating our base bounds.

Leaving this as a Draft PR since we probably want to wait until the upstream packages are published.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,1 +1,11 @@
 packages: tree-sitter tree-sitter-*
+
+source-repository-package
+  type: git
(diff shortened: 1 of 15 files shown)


XVilka:
Should be rebased - there are conflicts now.

Kleidukos:
There are many conflicts and we are four years after the fact now, I will take the liberty of closing this PR.

Issue #184, opened 2019-08-18
Tree suppose to be Storable?


IlhamKhabibullin:
Can't peek Tree from Ptr since Tree does not implement Storable typeclass. Is there some other proper way to do it?
I am pretty new in Haskell so it might be my fail. Sorry if it is.

robrix:
If I recall correctly, all of the API which uses `Tree`s actually takes `Ptr Tree`. Thus, there’s no real value in copying a `Tree` using `Storable` since the only thing you could do with it is write it back out to C memory using `poke` and then pass the resulting pointer around to existing API.

If you describe what you’re trying to accomplish with the `Tree`, we can probably advise you on the best way to use the API.

IlhamKhabibullin:
I wanted to print it, to see how the whole `Tree` is look like. Right now i have discovered that i suppose to do it with root node, but i can't find function called `ts_node_string` which is provided with original library.

IlhamKhabibullin:
I wanted to print it, to see how the whole Tree is look like. Right now i have discovered that i suppose to do it with root node, but i can't find function called ts_node_string which is provided with original library.

robrix:
@IlhamKhabibullin: Yes, we don’t currently bridge that function; I’ve no objection to doing so, however!

robrix:
Edited the issue title to note that `ts_node_string` is not currently bridged.

PR #292, opened 2020-09-11
Queries

Gauteab:
I think Queries are working properly.

This is the first time did anything with the FFI, so there are a few things I am unsure about.

I moved `Struct` into a new module to use it in Query. This should maybe be moved into an `Internal` directory.
I don't fully understand `Storable`, so I left some things unfinished there. Particularly `TSQueryError` which is currently just an Int.

While testing a bridged `ts_node_string` before seeing there was already an open [issue](https://github.com/tree-sitter/haskell-tree-sitter/issues/184). I also made a helper to call it on a `Node`.

Finally, I ran clang format (automatically) with WebKit styling on bridge.c. If it was using a particular formatter I can revert that.

### DIFF
--- a/tree-sitter-haskell/examples/Demo.hs
+++ b/tree-sitter-haskell/examples/Demo.hs
@@ -1,96 +1,80 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
-module Main (main) where
+module Main where
 
-import qualified Data.Tree as Tree
 import           TreeSitter.Parser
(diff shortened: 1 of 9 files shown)


dcreager:
Hi @Gauteab, thanks for the contribution!

> Finally, I ran clang format (automatically) with WebKit styling on bridge.c. If it was using a particular formatter I can revert that.

Yes, please revert the styling changes, since they're not a substantive part of the PR.  It's much harder for us to review if the diff contains a bunch of cosmetic changes to unrelated parts of the file.

PR #301, opened 2022-06-20
Update link to Haskell demo

robertoaloi:


robertoaloi:
Just realized the same fix was proposed in #299, so closing as a duplicate.

PR #321, opened 2024-09-21
Add ts_tree_edit

crumbtoo:
This currently pulls in `{eval,peek,poke}Struct` from `TreeSitter.Node`. Perhaps these should be moved to some `TreeSitter.Internal` module?


crumbtoo:
@Kleidukos

Kleidukos:
Excellent. I see this is binding to a new C function through FFI. Does this require a minimum version for the tree-sitter library?

crumbtoo:
As far as I'm aware, `ts_tree_edit` has been there since the [first release](https://github.com/tree-sitter/tree-sitter/blob/652eb3bbb62cf883a05c23a12e50df6caf92a45b/lib/include/tree_sitter/api.h#L102).

tree-sitter/src/TreeSitter/Tree.hs
```diff
@@ -16,5 +20,35 @@ withRootNode tree action = alloca $ \ ptr -> do
   ts_tree_root_node_p tree ptr
   action ptr
 
+data TSInputEdit = TSInputEdit
```
Kleidukos, line 23: Little nitpick: Could you please add some form of documentation to this struct? Does it reflect something from the C library? If yes, could you perhaps point tot he

Kleidukos (approved):

Kleidukos:
@crumbtoo oh wait, if I may ask one more thing: Could you please add an entry in the [changelog](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/tree-sitter/ChangeLog.md)?

